PR #54, opened 2015-09-08
NKRO

RingoM:
Reimplemented NKRO keyboard

### DIFF
--- a/examples/NKROkeyboard/NKROkeyboard.ino
+++ b/examples/NKROkeyboard/NKROkeyboard.ino
@@ -1,0 +1,54 @@
+/*
+  Copyright (c) 2014-2015 NicoHood
+  See the readme for credit to other people.
+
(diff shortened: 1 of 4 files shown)


NicoHood:
Thanks. Will have a look at this in a moment, also your Teensy Keyboard comment seems like a good idea.

NicoHood:
Can you please remove the space in the .cpp filename and resolve the merge conflicts (sorry, my fault)?

I'll edit this post if you should change anything else. I can do this later too though.

NicoHood:
Added it here:
https://github.com/NicoHood/HID/commit/78870a82dd6034684d2549a5ea74363969262e87

The example sketch causes my arduino ide to hang a lot. I am not sure if the implementation is wrong or the ide cant handle so fast so much buttons. It would be nice if you could have a look. I tested on linux.

PR #72, opened 2016-01-17
Add a simple Processing RawHID example

DeqingSun:
This is much easier than using HIDAPI in C.


NicoHood:
Even tough I have not looked at the example closely, it would be nice to know how to use the Arduino on the other side. Do you also have to use the current example or do you have different test code for the arduino side?

DeqingSun:
https://github.com/NicoHood/HID/tree/master/examples/RawHID/RawHID
I used your code with an off shelf Arduino Micro Pro
The only change I made was using Pin 10 for button.

DeqingSun:
I did a workaround by add a leading zero on windows platform.

PR #76, opened 2016-02-27
Use OR assignment for NKRO modifier keys

Scyntrus:
The DefaultKeyboardAPI implementation uses OR assignment instead of basic assignment.


NicoHood:
Good catch. I must have missed this with the last patch.

NicoHood:
https://github.com/NicoHood/HID/releases/tag/2.4.3

PR #117, opened 2017-08-19
Add Surface Dial support

wind-rider:
I added Surface Dial support using the documentation at https://docs.microsoft.com/en-us/windows-hardware/design/component-guidelines/radial-controller-input-reports and https://docs.microsoft.com/en-us/windows-hardware/design/component-guidelines/radial-controller-sample-report-descriptors.

Running it on an Arduino Pro Micro makes Windows recognize it as a Wheel device and I can control apps with it.

Remarks:
 - I disabled support for x / y position and device width (it is excluded from the report and the report descriptor).
 - The rotation needs to be 15 bit according to the report descriptor. It should be a signed number. However I'm not sure if I'm doing the conversion right there and whether I'm using the correct data types.
 - "Output reports" are not implemented yet, as I don't know how to do that. This could be used to support [haptic feedback](https://docs.microsoft.com/en-us/windows-hardware/design/component-guidelines/radial-controller-output-reports)


NicoHood:
I am unable to test this currently, as I have no hardware, nor am I using windows. Can someone else maybe verify if this is also working on his machine?

The example has no head comment. Maybe you can add it similar to the other examples. I cannot really test if the example is useful nor do I have the time.

I am willing to merge this, if you say it would also help other people. The rest of the code looks straight forward like a simple copy of mine, which should not have much more bugs as before :D

wind-rider:
@NicoHood it's a pity that you can't try it, anyway I added comments and a connection diagram.

wind-rider:
@NicoHood do you have any feedback on this?

NicoHood:
I just merged it, as I have no time to review. It will work or not. I think you made a good job here, so everything should be fine. Once I got more time, I will recheck all usb stuff again. Thanks for your contribution. :)

wind-rider:
Thanks for merging then :)

Issue #120, opened 2017-10-22
DefaultKeyboardAPI::set can leave KeyReport in a bad state


SteveBenz:
My PS2->USB keyboard translator manages to highlight this problem, which has been around for a while.

If you do:

press(key1)
press (key2)
release(key1)
press(key2)

Then DefaultKeyboarAPI::set fails to grok that key2 is already in the report, and causes key2 to appear in the key report twice.  So the next call to release(key2) will not work.

SteveBenz:
A pull request will follow shortly...

NicoHood:
Dont we need to check/fix the pressing code then? Only add a key, if its not in the list. Your code would also fix the issue, but from my point of view it would be better to fix it in the press() code, not release(). Correct me if I am wrong.

The code would work like this:
Search an empty slot, remember the number. Still check for the same key, abort if the key was already found. If not, save the key in the first empty slot.

SteveBenz:
There are two good ways to fix it.

1. You could make all the code paths look at all 8 bytes of the record (what you're proposing)
2. You could make all the code paths look at up to 8 bytes, stopping at the first empty entry (will result in the fastest code)

My change enables the second choice, but doesn't implement it, because I've always felt like the best policy in creating a first pull request into somebody else's codebase is to start out with the minimal change, rather than what I think is the best one (because there might be more I'm not aware of).

I'm happy to implement either of the two choices above; up to you.

SteveBenz:
I updated it to always keep the list without duplicates or 0's in the middle.  Basically, this means that the routine's speed depends on the number of keys pressed, which is good since most of the time that number is zero or one.

Tests are here:

https://github.com/SteveBenz/HidKeyboardSetTests

SteveBenz:
Not sure if you're busy with other things or what.  This is an important issue for any keyboard-based project.  Imagine a keyboard being used, say, for a WASD-based game - there will be plenty of times when a key will be held down while another key is being repeatedly pressed and released.  I found the issue in my application, which is a tweakable PS2-to-usb keyboard adapter.  There it primarily showed up because the PS2 keyboard has that auto-repeat feature, so it's reproducible just by holding down a pair of keys and releasing them in the order you pressed them.  ('a'<hold>'s'<hold> release a, release s, and now 's' is a dead key).

If you really wanted to follow the approach I outlined above, I created a second pull request that goes that road, but I think the first proposal is better just because it's fewer machine cycles.

NicoHood:
[...]

PR #121, opened 2017-10-22
Fixed issue #120 - unsetting a key can leave keyreport in a bad state

SteveBenz:


NicoHood:
Any reason why you opened two PRs?
Which one do you prefer and why?

src/HID-APIs/DefaultKeyboardAPI.hpp
```diff
@@ -24,46 +24,56 @@ THE SOFTWARE.
 // Include guard
 #pragma once
 
+DefaultKeyboardAPI::DefaultKeyboardAPI()
+{
+	memset(&_keyReport, 0, sizeof(_keyReport));
```
NicoHood: As far as I know this is not required, as c++ should init all variable members to zero. Right?

SteveBenz:
As I said in the other review, I think this is the better one.  The other one leaves keystrokes in the same position in the array, and perhaps there's compatibility reasons why that's good.  (Though I don't believe so).  This one is just faster -- in the most common cases, there will be at most one key pressed, and this algorithm will go through the loop at most two times for those cases.  The other implementation will go through the loop 6 times regardless of how many keys are actually pressed.

SteveBenz:
I closed the other review, because I think it's distracting and this is the way to go.  I do have a couple of questions:

1. Making 'set' return true if a key has already been pressed means that we'll end up sending a new report, even when we haven't changed the array at all.  This doesn't seem like a good thing to me.
2. I was testing holding down all 6 available keys, and I found that, even though I know the code is correct in its present form, it appears as if only 5 characters are coming across.  I say that because if I take a plain old keyboard and I 'press asdfjkl' (and hold down the letters as I go), I see it autorepeat on k, the 6th key.  On my keyboard built with HID, it autorepeats on the j, the 5th key.

It's not a big deal, but it's irksome.  I feel like there's a problem somewhere else.

SteveBenz:
(just to be clear, that 5th vs 6th character autorepeat happens the same with or without my changes).  I wonder if there's a descriptor somewhere that's off by one.

NicoHood:
This PR is very confusing to read, but I understand what it does. I have a few questions>
* Compared to the other PR, how much do they differ in size?
* If you reorder the keys on release, isnt this still as same slow as before?
* I think it must be possible to write this code with normal uint8_t i as index, not with pointers. This should make it better to read.

Please open another issue for the 5key thing that you notices. Also please write the OS that you are using and the arduino ide version.

SteveBenz:
**Quick summary** 
The old pull request (my [alternateFix branch](https://github.com/SteveBenz/HID/blob/alternateFix/src/HID-APIs/DefaultKeyboardAPI.hpp)) is 2 just bytes bigger than what's in your master branch right now.  The code in this branch is 16 bytes bigger.  However, a single press-release cycle (the most common case) is two times slower with alternateFix than with master.  Almost all of the win comes from the fact that _press_ always has to loop through the entire array if you allow for holes in the array, but if you don't allow holes, you can always stop at the first zero.  [Converting to arrays](https://github.com/SteveBenz/HID/blob/useIndices/src/HID-APIs/DefaultKeyboardAPI.hpp) significantly increased the code size and slowed performance and I don't think it did anything at all to help readability.

**Dance Mix:**
First, in poking around, I discovered I had an off-by-one error that had a surprisingly large effect on code size and perf.  These numbers reflect the new commit.

Regarding size, I would love to see the assembly, but from what I read, it's a bit of a project to get at it, so I'll settle for the size announced by the compiler.  With the change that we have in this pull request, it grows the code by 16 bytes.  For grins, I swapped out all the pointers for indices (see [line 26](https://github.com/SteveBenz/HID/blob/useIndices/src/HID-APIs/DefaultKeyboardAPI.hpp) and recompiled that.  I got 42 more than the baseline.  So, then I went back to the alternateFix branch (which uses indices, but always loops through the entire array.  To my surprise, that was only 2 bytes more than the baseline.  But hey, given my luck with pointers before, maybe that'd help here?  But nope, after converting it to pointers the code size actually grew by 4 bytes.  Moral of the story: Never try to outguess the C++ compiler.

As to the performance - the difference between master and alternateFix is that alternateFix leaves holes in the list of keys, so _press_ has to iterate through all 6 keys to be correct, always.  Master does not leave those holes, so in both _press_ and _release_ it can stop at the first empty slot.  So, it's natural to expect it to be quicker.  But bigger code is usually slower than smaller code so a test is in order.

I set up [this test](https://github.com/SteveBenz/PS2KeyboardHost/commit/d01287f3feec0618aa07ae7360ebada67af9c3cd).  With "master" it took spewed 9040ms and with the alternateFix it took 15112ms.  Accounting for overhead, it means alternateFix takes ~14us/press-release cycle and with master it's 7.5us/press-release.

As to readability of pointers vs. indices, it really depends a lot on who you are, right?  If you come from an stl background, where iterators are essentially pointers, then I think you'll be more comfortable with pointers.  There are legends that pointers are "hard".  I honestly believe that legend has its roots in Fortran, where there were no pointers.  CompSci professors of old weren't comfortable using them, and that carried on to their students & students' students.  But today, I work with many recent grads, and they just don't struggle with it.  (Although I should say that the company I work for can afford to be very picky.)

If you are really serious about readability, I'd suggest the way to go would be to split 'set' into separate press/release methods.  That would make a big difference in readability at a modest cost to code size.  Another good opportunity would be to meld DefaultKeyboard with DefaultKeyboardAPI and lose the fancy 'using' statements in DefaultKeyboardAPI.

NicoHood:
Closing as #148 fixed it already

PR #124, opened 2017-11-04
Fixed issue #120 - unsetting a key can leave keyreport in a bad state, resulting in stuck keys

SteveBenz:
Fixes the issue by always scanning the whole array

### DIFF
--- a/src/HID-APIs/DefaultKeyboardAPI.hpp
+++ b/src/HID-APIs/DefaultKeyboardAPI.hpp
@@ -47,9 +47,9 @@ size_t DefaultKeyboardAPI::set(KeyboardKeycode k, bool s)
 		// and if there is an empty slot. Remove the first available key.
 		uint8_t emptySlot = numKeycodeSlots;
 		uint8_t foundSlot = numKeycodeSlots;
-		for (uint8_t i = 0; i < numKeycodeSlots; i++)
+		for (uint8_t i = numKeycodeSlots; i > 0; --i)
 		{
-			auto key = _keyReport.keycodes[i];
+			auto key = _keyReport.keycodes[i-1];
 			if (key == KEY_RESERVED) {
 				emptySlot = i;
 			}
@@ -60,16 +60,11 @@ size_t DefaultKeyboardAPI::set(KeyboardKeycode k, bool s)
 
 		if (s && foundSlot == numKeycodeSlots && emptySlot < numKeycodeSlots) {
 			_keyReport.keycodes[emptySlot] = k;
-			return 1;
 		}
 		else if (!s && foundSlot < numKeycodeSlots) {
 			_keyReport.keycodes[foundSlot] = KEY_RESERVED;
-			return 1;
 		}
-		else {
-			// No empty/pressed key was found
-			return 0;
-		}
+		return foundSlot < numKeycodeSlots || (s && emptySlot < numKeycodeSlots);
 	}
 }
 
--- a/src/HID-APIs/DefaultKeyboardAPI.hpp
+++ b/src/HID-APIs/DefaultKeyboardAPI.hpp
@@ -42,7 +42,7 @@ size_t DefaultKeyboardAPI::set(KeyboardKeycode k, bool s)
 	}
 	// Its a normal key
 	else{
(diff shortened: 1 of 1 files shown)


src/HID-APIs/DefaultKeyboardAPI.hpp
```diff
@@ -42,28 +42,35 @@ size_t DefaultKeyboardAPI::set(KeyboardKeycode k, bool s)
...
+            }
+            else if (key == k) {
+                foundSlot = i;
+            }
+        }
+
+        if (s && foundSlot == numKeycodeSlots && emptySlot < numKeycodeSlots) {
+            _keyReport.keycodes[emptySlot] = k;
+            return 1;
+        }
```
NicoHood: If the key was already pressed, we should also return 1. If I am not completely mistaken the code could look like this:
```cpp
if (s && foundSlot < numKeycodeSlots) 
{    
    return 1;
}
else if (s && foundSlot == numKeycodeSlots && emptySlot < numKeycodeSlots) 
{
    _keyReport.keycodes[emptySlot] = k;
    return 1;
}
else ...
```

I am not sure what would be the best ordering of if/else logic here.
Can you please test this and suggest a better solution, if you have any?

src/HID-APIs/DefaultKeyboardAPI.hpp
```diff
@@ -42,28 +42,35 @@ size_t DefaultKeyboardAPI::set(KeyboardKeycode k, bool s)
...
-			
-			// Test the key report to see if k is present. Clear it if it exists.
-			if (!s && (key == k)) {
-				_keyReport.keycodes[i] = KEY_RESERVED;
-				return 1;
-			}
-		}
+        uint8_t emptySlot = numKeycodeSlots;
+        uint8_t foundSlot = numKeycodeSlots;
+        for (uint8_t i = 0; i < numKeycodeSlots; i++)
```
NicoHood: In this case the new key will be added to the end of the report, not the beginning. If counting downwards has no disadvantage (in code size), it could make more sense to do it this way. This is how other implementations work, I just try to avoid other problems, especially with older/bios systems. But since you tested this, it should be okay, also from the spec. Just as a personal thought...

src/HID-APIs/DefaultKeyboardAPI.hpp
```diff
@@ -42,28 +42,35 @@ size_t DefaultKeyboardAPI::set(KeyboardKeycode k, bool s)
 	}
 	// Its a normal key
 	else{
+        const int numKeycodeSlots = sizeof(_keyReport.keycodes) / sizeof(_keyReport.keycodes[0]);
```
NicoHood: Please use uint8_t here, this could result in smaller code size.

src/HID-APIs/DefaultKeyboardAPI.hpp
```diff
@@ -42,28 +42,35 @@ size_t DefaultKeyboardAPI::set(KeyboardKeycode k, bool s)
...
+            return 1;
+        }
+        else if (!s && foundSlot < numKeycodeSlots) {
+            _keyReport.keycodes[foundSlot] = KEY_RESERVED;
+            return 1;
+        }
+        else {
+            // No empty/pressed key was found
+            return 0;
+        }
```
NicoHood: Please fix the indention. It looks like the code was written with 4 spaces as indention, at least that is how I code nowadays.

SteveBenz:
I fixed the issues you raised.  I don't like this solution as much as I do the other one, because this one requires more machine instructions most of the time.  It loops through the whole array.  I just created it because I thought (based on something you had said before and the silence between) that perhaps there was a USB-specific reason to not move the keys around in the array.

This solution, by the way, is not as well tested as the other one - I created some unit tests for the other one and I run the other one in my keyboard at work.  I have a custom PS/2 to USB keyboard converter that also remaps some keys for me.  Alas.  I managed to leave both of my Arduinos at the office today, so I haven't actually tested this solution in this latest iteration in other than artificial scenarios.

SteveBenz:
Abandoning this avenue, as I don't think there's any compelling reason to choose this approach over the other.

PR #168, opened 2019-03-17
Added generic keyboard layout support

wladimir-computin:
Hey Nico,
your HID project is great, but it was missing support for my beloved german keyboard layout.
I added some sort of generic layout support. Of course I also added the german layout.
Code is tested and works with BootKeyboard and Kubuntu 18.10.


src/KeyboardLayouts/ImprovedKeylayouts.h
```diff
@@ -24,11 +24,59 @@ THE SOFTWARE.
 // Include guard
 #pragma once
 
-#if !defined(LAYOUT_US_ENGLISH)
-#error This API does not support non US english layouts.
+// Keyboard Modifiers
+enum KeyboardMods : uint16_t {
+	MOD_LEFT_CTRL		= 0b0000000100000000,
```
NicoHood: Using (1<<8) etc would be better to me.

src/HID-APIs/KeyboardAPI.hpp
```diff
@@ -154,14 +154,39 @@ size_t KeyboardAPI::set(uint8_t k, bool s){
 	}
 
 	// Read key from ascii lookup table
-	k = pgm_read_byte(_asciimap + k);
-	auto ret = set(KeyboardKeycode(k & ~SHIFT), s);
+	uint16_t key = pgm_read_word(_asciimap + k);
```
NicoHood: This of course adds some more boilerplate for non us layouts. That is what I was trying to avoid. Just saying. Maybe using another wrapper could help still having a compact us layout and larger german layout at the same time.

src/HID-Settings.h
```diff
@@ -36,7 +36,7 @@ THE SOFTWARE.
 //#define LAYOUT_FRENCH
 //#define LAYOUT_FRENCH_BELGIAN
 //#define LAYOUT_FRENCH_SWISS
-//#define LAYOUT_GERMAN
+#define LAYOUT_GERMAN
```
NicoHood: The default must be US of course.

src/KeyboardLayouts/ImprovedKeylayouts.h
```diff
@@ -490,21 +538,10 @@ enum KeyboardKeycode : uint8_t {
     HID_KEYBOARD_RIGHT_ALT	= 0xE6,
     HID_KEYBOARD_RIGHT_GUI	= 0xE7,
 };
+*/
 
-// Keyboard Leds
-enum KeyboardLeds : uint8_t {
```
NicoHood: Why did you remove those?

src/KeyboardLayouts/ImprovedKeylayouts.h
wladimir-computin: Yeah, me too. But (1>>15) throws a compile error complaining the resulting number does not fit into uint16_t. It was late, so I went with the lazy solution...

NicoHood:
Thanks for taking the time. I gave it a quick review. Looks good overall.

I was thinking of keeping the existing code, but adding another wrapper around the keycode api to support additional layouts. This way you would still have a compact german keyboard and you can even provide additional layouts as extra library, it does not have to get merged in this project (but its still the best place though.). Do you understand what I mean?

I am on vacation for more than a week, so please dont wonder if I dont respond that soon again.

src/HID-APIs/KeyboardAPI.hpp
wladimir-computin: I don't think so. Reading 2 instead of 1 Byte once every Keystroke won't slow down anything, at least compared to the rest of the code. A few more comparisons for ALT, CTRL etc. are also neglible on those arduinos.

src/HID-Settings.h
wladimir-computin: Yep, you're right!

NicoHood: Could you please fix this? Is there also a way to change the layout without editing the source files? It is a shame, that arduino still has no option to define variables like this inside the user sketch.

src/KeyboardLayouts/ImprovedKeylayouts.h
NicoHood: I guess you mean 1<<15 not 1>>15? Maybe you need to add a cast?

src/KeyboardLayouts/ImprovedKeylayouts.h
```diff
@@ -24,11 +24,59 @@ THE SOFTWARE.
...
+	LED_SCROLL_LOCK		= (1 << 2),
+	LED_COMPOSE			= (1 << 3),
+	LED_KANA			= (1 << 4),
+	LED_POWER			= (1 << 5),
+	LED_SHIFT			= (1 << 6),
+	LED_DO_NOT_DISTURB	= (1 << 7),
+};
+
+#if defined(LAYOUT_US_ENGLISH)
+	//#include "ImprovedKeylayoutsUS.h"
```
NicoHood: Why is this commented? Does US layout still work?

NicoHood:
As I currently have no time myself to work on this more than checking the comments, I am happy to merge your PR. Please verify that US layout works and it is still set as default. Then I will merge.

Of course you can mention me, why not :-D Send me the link/name of magazine or even a copy (also living in germany). I am Happy to check it out!

src/KeyboardLayouts/ImprovedKeylayouts.h
wladimir-computin: I commented it out while debugging and then forgot about it.
US layout now works as expected.

src/HID-Settings.h
wladimir-computin: As you said, it's not possible the easy way, at least according to http://forum.arduino.cc/index.php?topic=262101.0.

A constructor-based approach is possible, but also requires a rather major rewrite and could cause some overhead. I guess HID-Settings.h is fine for now?

src/KeyboardLayouts/ImprovedKeylayouts.h
wladimir-computin: Yeah, cast was needed. Fixed.

wladimir-computin:
I finally found time to push the changes you requested. Feel free to merge now!
The magazine is c't, I will contact you once the article is ready :)

NicoHood:
Thank you for the changes! I will tag a new release.

Issue #167, opened 2019-03-10
what is the position data in the surface dial report for?


OmegaRogue:

NicoHood:
@hansmbakker could you please answer? He wrote that code.

hansmbakker:
You can put the official surface device on the surface studio screen and then windows will draw the surface dial menu around the device.

In the protocol it has this meaning:
- if you set it, it lets windows know where on the screen you placed your dial device
- if it is not set, Windows thinks the dial device is not placed on the screen but e.g. on the desk.

Please refer to the [official docs](https://docs.microsoft.com/en-us/windows-hardware/design/component-guidelines/radial-controller-protocol-implementation) for more info. Also note that I did *not* implement the position part of the protocol (it is optional), but only the button and rotation properties (which are required).

hansmbakker:
Feel free to submit a pull request if you implement it, though :)

OmegaRogue:
but how does the surface dial know where it is placed?

OmegaRogue:
and im implementing it

OmegaRogue:
> and im implementing it

https://github.com/NicoHood/HID/pull/169

hansmbakker:
> but how does the surface dial know where it is placed?
I have no idea. [This pdf](https://news.microsoft.com/windows-event-2016-assets/Surface-Dial-Fact-Sheet.pdf) says that it is done by PixelSense technology (which is just the name of the surface screens, so it gives no explanation).

https://docs.microsoft.com/en-us/windows-hardware/design/component-guidelines/radial-controller-designs says that it is vendor-specific and also gives no details.

[This video](https://youtu.be/JzqLCIwIJYQ?t=71) at 1:11 shows a digitizer plate. I suspect there is an (active) matrix in the Surface screens that the Dial can detect, just like a stylus could work?

hansmbakker:
By the way, what are you planning to use the position properties, and the Dial code in general, for?

OmegaRogue:
> By the way, what are you planning to use the position properties, and the Dial code in general, for?

A diy surface dial like hid device

hansmbakker:
Nice, good luck :)

OmegaRogue:
what are the indentation guidelines for the code? tabs or 4 spaces?

hansmbakker:
This is a question for @nicohood, I just meant to ask to make it consistent.

OmegaRogue:
because in some files its 2 spaces, in some its tabs

[...]

PR #170, opened 2019-04-24
added editorconfig as seen in #167

OmegaRogue:


OmegaRogue:
#167

NicoHood:
I dont think this is required, as it is a config for your personal editor, that I or other people do not use.

OmegaRogue:
editorconfig is cross editor https://editorconfig.org/

.editorconfig
```diff
@@ -0,0 +1,10 @@
+[*]
+charset = utf-8
+end_of_line = crlf
```
NicoHood, line 3: I guess this is wrong, isnt it? I'd prefer the unix linefeed. If some files are not yet in that format, that might be true, as they are very old.

.gitignore
```diff
@@ -0,0 +1,55 @@
...
+
+# Crashlytics plugin (for Android Studio and IntelliJ)
+com_crashlytics_export_strings.xml
+crashlytics.properties
+crashlytics-build.properties
+fabric.properties
+
+# Editor-based Rest Client
+.idea/httpRequests
+
```
NicoHood, line 55: I think this one is a bit oversized, I dont want to add this "bloat" to every project. Correct me, if I am wrong.

NicoHood:
I revisited this, and you are right. I totally missed the point with editor config. I have no much experience with this, but adding a basic config (with spaces!!!) is fine for me. Checkout my comments.

NicoHood:
Fixed with https://github.com/NicoHood/HID/commit/b179743ba750d493c2bfbe892c6c0d7104943e6a

Issue #175, opened 2019-06-25
On switching language specific input method


blahlicus:
@wladimir-computin I see you are the author behind the generic input method support, would you mind providing some insight on how to use this feature?

Suppose I want to use the DE layout which you have provided, do I `#define LAYOUT_GERMAN` in my Arduino sketch or will I have to edit HID-Settings.h in the library itself?

Also, I see in your HID-Settings.h that you have the following:

```
//================================================================================
// Settings
//================================================================================

#define LAYOUT_US_ENGLISH
//#define LAYOUT_CANADIAN_FRENCH
//#define LAYOUT_CANADIAN_MULTILINGUAL
//#define LAYOUT_DANISH
//#define LAYOUT_FINNISH
//#define LAYOUT_FRENCH
//#define LAYOUT_FRENCH_BELGIAN
//#define LAYOUT_FRENCH_SWISS
//define LAYOUT_GERMAN
//#define LAYOUT_GERMAN_MAC
//#define LAYOUT_GERMAN_SWISS
//#define LAYOUT_ICELANDIC
//#define LAYOUT_IRISH
//#define LAYOUT_ITALIAN
//#define LAYOUT_NORWEGIAN
//#define LAYOUT_PORTUGUESE
//#define LAYOUT_PORTUGUESE_BRAZILIAN
//#define LAYOUT_SPANISH
//#define LAYOUT_SPANISH_LATIN_AMERICA
//#define LAYOUT_SWEDISH
//#define LAYOUT_TURKISH
//#define LAYOUT_UNITED_KINGDOM
//#define LAYOUT_US_INTERNATIONAL

//================================================================================
// Definitions and Helpers
//================================================================================

// Default US keyboard layout
#if !defined(LAYOUT_CANADIAN_FRENCH) && !defined(LAYOUT_CANADIAN_MULTILINGUAL) \
&& !defined(LAYOUT_DANISH) && !defined(LAYOUT_FINNISH) && !defined(LAYOUT_FRENCH) \
&& !defined(LAYOUT_FRENCH_BELGIAN) && !defined(LAYOUT_FRENCH_SWISS) && !defined(LAYOUT_GERMAN) \
&& !defined(LAYOUT_GERMAN_MAC) && !defined(LAYOUT_GERMAN_SWISS) && !defined(LAYOUT_ICELANDIC) \
[...]

PR #176, opened 2019-06-28
Allow sketches to change the keyboard language layout config without editing source files in the library itself

blahlicus:
As per #175, NicoHood/HID currently requires the user to change the flags in `HID-Settings.h` to switch from US to German keyboard layout and etc.

This is really bad because it requires the user to edit source files in the HID-Project library.

This PR allows sketches to set flags to define which layout it would like to use without messing with source files in the library, below is an example using the German layout:

```
#define HID_CUSTOM_LAYOUT // set this flag to indicate that a custom layout is selected
// if the above flag is not defined, then the default US layout is used instead

#define LAYOUT_GERMAN // set this flag after the above flag to indicate the custom input method is German
// for more layouts, see /src/KeyboardLayouts/ImprovedKeylayouts.h

// you must set the above flags before including HID-Project
#include "HID-Project.h"


void setup() {
  BootKeyboard.begin();
  delay(300);
  BootKeyboard.write(KEY_7|MOD_LEFT_SHIFT); // this outputs the / symbol as per the german input method
}


void loop() {

}
```

To use a non-US English layout, the user only has to `#define HID_CUSTOM_LAYOUT` followed by `#define LAYOUT_[Your layout here]` before `#including` the `HID-Project.h` file.

if `HID_CUSTOM_LAYOUT` is undefined, then it defaults to the original US ASCII table, which means this PR is reverse compatible with older sketches.


wladimir-computin:
How do you compile it? Are you using the Arduino IDE?

blahlicus:
Yeah.

NicoHood:
Oh nice. This works, because I did not use .cpp files, right? The .cpp files do not require the keyboard layout. This should work - I am surprised that it is that easy. Did you test this carefully?

blahlicus:
I tested the above code on both the Pro Micro and an SAMD21 device.

I also tested it in my complicated firmware which depends on this library and it worked fine. Previous version of my complicated firmware which depends on older versions of this library also works with this so it shows reverse-compatibility is working.

Perhaps we could have a second opinion from @wladimir-computin and then decide on the merge?

wladimir-computin:
In my opinion @blahlicus did a great job, no complaints from my side.
@NicoHood, are you ok with merging his commit?

bitboy85:
Not working for me https://github.com/NicoHood/HID/issues/208
Arduino IDE 1.8.11

PR #190, opened 2019-09-23
Starting development for 3D mouse (Multi-axis controller)

Matesaktesak:
Here is the HID descriptor as specified by https://www.usb.org/sites/default/files/documents/hut1_12v2.pdf . Alrady working on the real code for its function, but I'm not a great coder so it may be full of errors...


Matesaktesak:
I have a problem where windows say that my descriptor is not byte aligned - can you help me fix it?

NicoHood:
new pr: https://github.com/NicoHood/HID/pull/226

Issue #194, opened 2019-10-08
Cant get gamepad examples working on SAMD21


DCooper-nz:
I am trying to get the gamepad code working on a SAMD21 board and am having no joy. The code compiles and uploads fine, the board is identified in the windows usb controller setup but I get no data from the board. The same code works fine on a Leonardo. Also the keyboard and mouse examples both work fine on the SAMD21 just not the gamepad. I have tried multiple SAMD21 based boards and several computers with the same result.

thomasgauthier:
I have the same issue on the Adafruit Feather M0. Arduino Leonardo works fine.

jrowberg:
Same symptoms here. Could this be due to the relatively recent modifications to the USB implementation in the samd core (at least Adafruit’s fork, I think the official Arduino repo as well) which allows switching in the IDE to TinyUSB instead? I see the game pad with many axes and a zillion buttons, but nothing ever changes when advancing states via the button/interrupt.

EtienneGameSeed:
Same here, no refresh. Device appears as expected, but no report is received.

gdsports:
@DCooper-nz @thomasgauthier @jrowberg @EtienneGameSeed The gamepad works with this patch. The solution is the union/struct for the HID report must be packed to avoid the addition of extra filler bytes. Tested with Gamepad, Gamepad1, Gamepad2, Gamepad3, and Gamepad4 on Ubuntu 18.04 with jstest-gtk.

```
diff --git a/src/HID-APIs/GamepadAPI.h b/src/HID-APIs/GamepadAPI.h
index 6a50018..0a9fe4a 100644
--- a/src/HID-APIs/GamepadAPI.h
+++ b/src/HID-APIs/GamepadAPI.h
@@ -38,15 +38,16 @@ THE SOFTWARE.
 #define GAMEPAD_DPAD_LEFT 7
 #define GAMEPAD_DPAD_UP_LEFT 8
 
+#define ATTRIBUTE_PACKED  __attribute__((packed, aligned(1)))
 
-typedef union {
+typedef union ATTRIBUTE_PACKED {
        // 32 Buttons, 6 Axis, 2 D-Pads
        uint8_t whole8[0];
        uint16_t whole16[0];
        uint32_t whole32[0];
        uint32_t buttons;
 
-       struct{
+       struct ATTRIBUTE_PACKED {
                uint8_t button1 : 1;
                uint8_t button2 : 1;
                uint8_t button3 : 1;


```

[...]

PR #211, opened 2020-03-02
Fix alignment for SAMD

gdsports:


NicoHood:
Did you make sure this still works on avr?

NicoHood:
Can you please also adapt the changes to the other files? They all require this patch it seems
https://github.com/NicoHood/HID/issues/194#issuecomment-667607045

gdsports:
OK, I will fix for other devices and check to make sure it works on AVR. My plan is to add #define ATTRIBUTE_PACKED to HID-Settings.h. For AVR arch, the #define will do nothing so no change for AVR. For SAMD, it wll expand to force 1 byte alignment. Not sure if this is need for SAM (Due). I will check this also.

gdsports:
ATTRIBUTE_PACKED is defined in HID-Settings. For AVR, it does nothing. For SAMD and SAM, alignment is set to prevent the insertion of pad bytes.

The following examples tested on Linux Ubuntu 18.04 using an Arduino Zero (SAMD21). Not sure how to test raw HID and alternate keyboard. Examples also tested on an Arduino Leonardo.

- consumer
- gamepad
- absolute mouse
- improved mouse
- boot keyboard
- improved keyboard
- keyboard led
- nkrokeyboard

NicoHood:
Thanks!

PR #230, opened 2020-08-29
Add italian keymap

SukkoPera:
**NOTE: THIS PR IS NOT YET READY FOR MERGING**.

Here's the Italian keymap, tested and working correctly.

Although, I think there's something that needs to be discussed before this can be merged and more keymaps be made: the following keys do not exist in the IT layout:
```
	KEY_TILDE
	KEY_EQUAL
	KEY_RIGHT_BRACE
	KEY_LEFT_BRACE
	KEY_SEMICOLON
	KEY_SLASH
```
The symbols these keys should produce are obtained somehow else. So I checked what had been done in the DE layout and removed these altogether and added new keys matching the symbols we have in the IT layout (e.g.: `KEY_AGRAVE`, `KEY_PLUS`, etc.). Some other key names were moved (e.g.: out `KEY_MINUS` is the US `KEY_SLASH`).

But after some thinking I'm not sure this is the way to go, since it can create a number of problems:
   1. Any code using the removed keys will break when the keymap is changed.
   1. There's no way to press a specific key independently from the used keymap. How can I always press "the key next to 0", regardless of the used keymap?

So maybe **the key names should be intended as referring to the physical key of the US layout** they get the name from. This means they should never change. The individual keymaps could introduce some aliases, but **their main job would only be the customization of `_asciimap` using the key combos that would be used on a US keyboard**. So, for instance, referring to the IT layout, `KEY_MINUS` would still refer to scancode 45 and `_asciimap` would contain `KEY_SLASH` at the row corresponding to `-`.

Then **programs could use the `KEY_*` identifiers when they want to press a physical key, and just a *char* when they want to type that character**, i.e.:

    BootKeyboard,press (KEY_MINUS);

would always press "the key next to 0", whatever keymap is being used and

    BootKeyboard,press ('-');

would always enter a minus character, whatever keymap is being used.

For the latter to work best, **I have extended `_asciimap` beyond 127**, essentially turning it into a ISO-8859-15 map. This is nice since it allows a lot of language-specific characters and the Euro symbol to be entered. I'm not sure something like

    BootKeyboard,press ('à');

would work and be portable though, but one can just do:

    BootKeyboard,press (224);

Maybe we could introduce constants to make that easier (something like `CHAR_AGRAVE`).

Let me know what you think, I will then update my PR accordingly.


SukkoPera:
Well, I went ahead and made another PR with my proposal, see #231.

SukkoPera:
Superseded by #231, closing without merging.

Issue #242, opened 2020-12-21
Not compiling with Arduino SAMD boards version 1.8.10


wanthalf:
The library stopped compiling with Arduino SAMD Boards after upgrade to version 1.8.10 in the Boards Manager. The error is:

`Arduino/libraries/HID-Project/src/HID-Settings.h:107:10: fatal error: USB/PluggableUSB.h: No such file or directory
 #include "USB/PluggableUSB.h"`

It still compiles well with all versions up to 1.8.9., when downgraded. I am not sure if it is more a problem of the Arduino SAMD Boards libraries or this library.

(Tested in Arduino IDE 1.8.13 on macOS 11.1, compilation for Arduino Zero (Native USB Port)).

arunmjayan:
I am also facing the same problem. Did anyone overcome this issue, if yes, can you please share how?

NicoHood:
There is a fix: https://github.com/NicoHood/HID/pull/244

If anyone is willing to check avr support, I am happy to merge this.

PR #244, opened 2021-01-06
Compile fails with Arduino SAMD board package	1.8.10 and newer

gdsports:
Proposed fix for https://github.com/NicoHood/HID/issues/242

Compiles on the following SAMD board packages: Arduino, Adafruit, and Seeeduino.

### DIFF
--- a/src/HID-Settings.h
+++ b/src/HID-Settings.h
@@ -111,12 +111,17 @@ THE SOFTWARE.
 
 #define ATTRIBUTE_PACKED  __attribute__((packed, aligned(1)))
 
+#define USB_EP_SIZE                 EPX_SIZE
+#define EP_TYPE_INTERRUPT_IN        USB_ENDPOINT_TYPE_INTERRUPT | USB_ENDPOINT_IN(0);
+#define EP_TYPE_INTERRUPT_OUT       USB_ENDPOINT_TYPE_INTERRUPT | USB_ENDPOINT_OUT(0);
+
+#if defined(ARDUINO_API_VERSION)
+#include "api/PluggableUSB.h"
+#define EPTYPE_DESCRIPTOR_SIZE		unsigned int
+#else
 #include "USB/PluggableUSB.h"
 
 #define EPTYPE_DESCRIPTOR_SIZE      uint32_t
-#define EP_TYPE_INTERRUPT_IN        USB_ENDPOINT_TYPE_INTERRUPT | USB_ENDPOINT_IN(0);
-#define EP_TYPE_INTERRUPT_OUT       USB_ENDPOINT_TYPE_INTERRUPT | USB_ENDPOINT_OUT(0);
-#define USB_EP_SIZE                 EPX_SIZE
 //#define USB_SendControl           USBDevice.sendControl -> real C++ functions to take care of PGM overloading
 #define USB_Available               USBDevice.available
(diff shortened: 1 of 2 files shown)


src/HID-Settings.h
```diff
@@ -111,12 +111,17 @@ THE SOFTWARE.
 
 #define ATTRIBUTE_PACKED  __attribute__((packed, aligned(1)))
 
+#define USB_EP_SIZE                 EPX_SIZE
+#define EP_TYPE_INTERRUPT_IN        USB_ENDPOINT_TYPE_INTERRUPT | USB_ENDPOINT_IN(0);
+#define EP_TYPE_INTERRUPT_OUT       USB_ENDPOINT_TYPE_INTERRUPT | USB_ENDPOINT_OUT(0);
+
+#if defined(ARDUINO_API_VERSION)
+#include "api/PluggableUSB.h"
+#define EPTYPE_DESCRIPTOR_SIZE		unsigned int
```
NicoHood, line 120: why is it unsigned int here? If so, please use uint16_t

src/port/samd.cpp
```diff
@@ -2,12 +2,14 @@
 
 #ifdef ARDUINO_ARCH_SAMD
 
+#if !defined(ARDUINO_API_VERSION)
```
NicoHood, line 5: Why are those not required anymore? Are they implemented by the api?

src/HID-Settings.h
gdsports: These changes are based on similar changes the Arduino folks made to their MIDIUSB library having to do with ARDUINO_API. I suspect the idea is to have the same USB API for AVR and SAMD. unsigned int will be 16 bits for AVR and 32 bits for SAMD. I have no direct information on the purpose for the ARDUINO_API changes.

https://github.com/arduino-libraries/MIDIUSB/commit/80b6fac501702a268d7f9ab9702333c5d059e059

src/port/samd.cpp
gdsports: Yes, the SAMD board release now has USBAPI.h which include routines for USB_SendControl and others.

https://github.com/arduino/ArduinoCore-API/blob/master/api/USBAPI.h

As far as I can tell, the latest SAMD board package releases no longer match github.com/arduino/ArduinoCore-SAMD. It appears the files from ArduinoCore-API are included in the SAMD board packages during the release build procedure.

src/HID-Settings.h
NicoHood: Alright, that makes sense. Was not aware that its 32 bit on SAMD

src/port/samd.cpp
NicoHood: Would you mind finding this out for avr? It would be best to have a solution that works with every board, of course ;-)

NicoHood: Any chance to test the AVR code? I'd be happy to merge then ;-)

NicoHood:
@SukkoPera Maybe you can give this a regression test on AVR and see if it still works? That'd be nice, so we can include it in the next release along with your keylayout fixes.

SukkoPera:
Everything compiles fine for ATmega32U4. I did a quick 5 minute test with SingleGamepad and BootKeyboard and everything seems to be fine. Otherwise you wouldn't be reading this :).

Issue #133, opened 2018-04-16
Expected to work?


MartinYoungOB:
I've been trying to get raw HID. It's partially working but not quite. I'm wondering if it's expecting to work and if not, what I need to do to make it work? I've been looking through the source and there are quite a comments that suggest that it might not be complete. I don't understand many of them.

I have a genuine Arduino Uno with a 16u2. I've put Hoodloader2 on it and it works exactly as expected - I can upload programs to it and communicate with the other AVR on the same board.

I'm now trying to use RawHID to set up a bidirectional link between my Uno and the PC using my own vendor id and product id. I'm using hidapi on Windows 10 (from the main repo) on the PC.

When I enumerate devices on the PC, I see the 16u2 and I can connect to it. I can send bytes with hid_write() on the PC and receive them using RawHID.read() on the 16u2. This is great. (hidapi sends on the first OUT endpoint or the control endpoint is there isn't an OUT one. I'm not sure which it happening this case.)

However when I try to send from the 16u2 using RawHID.write() it is not received by the PC. Using Wireshark I can see that each character is arriving as the single payload byte in a URB_INTERRUPT packet but hid_read() does not return them.

Is this expected to work? Is there something else I need to do?

Also:

 - RawHID doesn't work in a multi-report interface. Because of the serial port that the 16u2 is implemented is this a multi-report system. Do I need to turn off the serial port somehow?

 - Can I induce RawHID to send on endpoint 0, the control endpoint?

Thanks,
Martin.

NicoHood:
RawHID support was always unstable. Also there were bugs introduced with the newer IDE versions which should be corrected now. But I've never retested it and its been quite a long time since I wrote the code.

You can simply just try it yourself and find the bugs. I sadly have no time to do so.

Multireport does not work with hidapi. It just confuses the API. But I am not using multireports, so nothing to do here.

RawHID receives data on a dedicated endpoint and sends data on endpoint0. You could send it on its own endpoint, but this library is not implemented like that.

NicoHood:
Your wireshark observation is correct and expected.

eivan7538:
Nico, Martin, did you have any success with Raw HID? I just tried it with the basic rawhid_test.c with Arduino MKR Zero and had the same issue: Arduino can receive packets from PC but when Arduino sends packets the PC does not receive anything. I would love to help but I'm an amateur so I'm not sure if I can debug and fix it.

zaphodbe:
[...]

PR #245, opened 2021-01-11
Update regarding the HID-Project.h

arunmjayan:
Hi there, 
We are thankful for your support, i came here to request you to include the RawHID.h file into the HID-Project.h

### DIFF
--- a/src/HID-Project.h
+++ b/src/HID-Project.h
@@ -52,5 +52,6 @@ THE SOFTWARE.
 #include "SingleReport/SingleNKROKeyboard.h"
 #include "MultiReport/NKROKeyboard.h"
 #include "MultiReport/SurfaceDial.h"
+#include "SingleReport/RawHID.h"
 
 // Include Teensy HID afterwards to overwrite key definitions if used


NicoHood:
I am not yet sure why I removed it. I guess because it is not working properly yet. There are several issue about that already open. If you find a way to fix those, I am happy to enable the code again.

spuder:
Here is one of the issues specifically https://github.com/NicoHood/HID/issues/133

Currently HID-Project is able to _receive_ data from a PC with `RawHID.read()` but it is not able to _send_ data back to the PC with `RawHID.write()`

NicoHood:
@spuder How can we fix raw hid? I really do not remember the current state. What is needed to make write work properly?

spuder:
I don't fully understand the bug, but as long as the RawHID sends an array of atleast 64 bytes it works fine. 
```cpp
    // Workaround for bug when sending less than 64 bytes of data
    for (byte i = 0; i < sizeof(rawhidData); i++)
    {
        rawhidData[i] = 0x00;
    }
    RawHID.begin(rawhidData, sizeof(rawhidData));
```

If you attempt to send an array smaller than 64 bytes, nothing is transferred. 

Here is a working example 

https://github.com/spuder/MuteMe-arduino/blob/main/src/main.cpp

NicoHood:
Yeah, I think this has to be because a zero length package is missing or so.

mcuee:
The current git seems to work reasonaby well based on my testing using the example here.
* https://github.com/todbot/hidapitester/issues/20

There is a bug for the Feature report but the fix is simple.
* https://github.com/NicoHood/HID/issues/418

PR #261, opened 2021-03-09
Add spanish keylayout

SukkoPera:
Add Spanish keyboard layout.

Note that:
- I have never seen a Spanish keyboard in my life, I did this based on pictures I found on Google. I decided to make this attempt because it did not look too different from the Italian keyboard.
- I have no idea of how many different Spanish layouts exist (Spain, Latin America, etc) and which one(s) this fits.
- I have no experience with dead keys, so I assumed a single tilde or accent cannot be entered without pressing Space afterwards. This means a single tilde or accent cannot be typed at all, the library currently cannot support this (Correct me if I'm wrong).

I think someone from Spain should test and improve this before it gets into mainline.


NicoHood:
Wow, that is great!
You've written an excellent testing script. How about adding a spanish keymap to your system and testing that out?

SukkoPera:
That would really be the last resort, I'm a bit afraid not to choose the right variant. Other than that, I'd rather work on creating other layouts.

PR #262, opened 2021-03-10
More keylayouts

SukkoPera:
Add Danish, Finnish, French, Belgian, French/German Swiss, UK/Irish, Norwegian, Portoguese and Swedish keylayouts.

The same caveats as #261 (which must be merged separately, by the way) apply: I have never used any of these keyboards, I only did these based on pictures and Wikipedia.

I have made sure that they all compile but they are a bit too many for me to test deeper at this point. I'd suggest just merging them in and waiting for bug reports, if any.


NicoHood:
Great!

> I'd suggest just merging them in and waiting for bug reports, if any.

That is at least honest :-D I thought the same actually XD

I will give those PRs a bit of time, maybe someone will come here and give it a short test. Thanks a lot for those fixes!

Issue #267, opened 2021-03-25
Keyboard.write(KEY_F13); misbehaves on windows?


el-germi:
the following code using the standard arduino keyboard library works as expected, sending an 'F13' keystroke, but using HID-Project.h causes 'h'.
On the 'ImprovedKeylayouts.h' line 172 it says that it's expected to not work but why tho, and is there a way to fix it whilst keeping the System.write(SYSTEM_SLEEP) and System.write(SYSTEM_WAKE_UP) functions?

using a 5v arduino pro micro(32U4) and no HID_CUSTOM_LAYOUT.
windows keyboard layout is Spanish qwerty

Am I doing something wrong?

NicoHood:
Can you post me a *minimal* example that demonstrates the issue? I'd expect that `Keyboard.write(KEY_F13)` should work.

el-germi:
I can't, it was entirely my fault
I have a simple de-bounce function `void type(int key) {
  Keyboard.write(key);
  delay(500);
}` but as it turns out, the keys are objects in an enum and not defines like in the original;
replacing it with `void type(KeyboardKeycode key) {
  Keyboard.write(key);
  delay(500);
}` fixed my issue.
I'm terribly sorry for wasting your time :c
I should've read the code before posting

PR #270, opened 2021-03-27
fix compatibility issues with default keyboard

el-germi:
A temporary solution to the compatibility problem between windows and your fancy keyboards. By checking if the default keyboard library was imported beforehand you can get back the "Keyboard.write(KEY_F13)" without loosing "System.write(SYSTEM_SLEEP)"
See "issue" #267 and the fact that importing both libraries (Keyboard.h and HID-Project.h) causes the keys to be redefined and crashes the compiler

### DIFF
--- a/src/HID-Project.h
+++ b/src/HID-Project.h
@@ -48,10 +48,13 @@ THE SOFTWARE.
 #include "MultiReport/System.h"
 #include "SingleReport/RawHID.h"
 #ifndef KEYBOARD_h
-#include "SingleReport/BootKeyboard.h"
-#include "MultiReport/ImprovedKeyboard.h"
-#include "SingleReport/SingleNKROKeyboard.h"
-#include "MultiReport/NKROKeyboard.h"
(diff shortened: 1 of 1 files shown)


NicoHood:
Instead of simply silently enabling the other library, I'd rather output an error message.

el-germi:
The main issue was the limitations of the layout based approach on windows.
My solution focuses more on that, enabling the classic keyboard without loosing too much functionality.
The redefinition error fix is just icing on the cake.

el-germi:
The latest commit adds a couple of "#pragma message", I don't know how descriptive they're supposed to be, but feel free to change them if you want

NicoHood:
I think the pragma should be replaced by an error message. I dont see any usecase where using the "normal" arduiono libraries combined with this one makes sense.

el-germi:
I use F13-F24 on windows, with the default library it's no problem at all, but never manage to do it using any of you keyboards (maybe it's my fault?) the bottom line is there are reasons to want the default library back without loosing everything

Issue #123, opened 2017-10-30
Capacitive Touch Screen Emulation


mbt28:
Hi,

Is it possible to emulate multi touch capacitive touch screen with this library?

Thanks

NicoHood:
I dont think so. You can only emulate an absolute mouse.

NicoHood:
Because I dont know how to have multiple inputs? Do you have any sample of a working HID device?

mbt28:
Actually yes, and there are device descriptors which can be found in those links. I am working on a simple library to make it with pro micro. Still it doesnt function, when it functions I will put on github.
(I am not windows fun but the explanation is quite good)
https://docs.microsoft.com/en-us/windows-hardware/design/component-guidelines/supporting-usages-in-digitizer-report-descriptors

With teensy:
https://www.codeproject.com/Articles/1001891/A-USB-HID-Keyboard-Mouse-Touchscreen-emulator-with

mbt28:
Here is my first trial code if you are interested, but windows says that "A request for the USB device descriptor failed." I couldnt try it with linux yet because now I only have my work pc.

https://gist.github.com/mbt28/406bdf15a248029c774085832c7c0c0c

amol-india:
Hi mbt28 ,
I tried your code on my arduino-leonardo +windows 7 os.
Windows recognized as single touch digitizer but it not worked further.
I checked under Usblyzer .I found following problem in your sketch.
You are sending 8 byte instead 7 byte in sendreport() function.
(first byte is Report Id and remaining 7byte.
I modified your code .Now windows recognizes as single touch digitizer.
I emulated simple click,double-click and moving cursor to selected x,y coordinate.

mbt28:
Hi

Actually I wrote usb multi touch hid to Arduino. But my code is such a mess now, working properly. Nowadays I have heavy workload once I have free time I will put it to github.

MahendraSondagar:
Hello **mbt28**

                  I'm also working on the same project, but with raspberry pi Zero-W. I have emulator keyboard via Bluetooth  and it's working great

 I also need to emulate HID mutitouch screen with raspberry pi Zero w  to send the data to host via bluetooth
Any one have reference with multi-touch screen please ?
Thanks

AiueoABC:
@mbt28 
I'm playing with your code. I confirmed windows can recognize this as a touch device, but not sure how to check it's working correctly.

If there is any progress, I want to know about it.
[...]

PR #297, opened 2021-06-20
Multi-touch touchscreen support (#123)

ilufang:
Hi, I was trying to implement an Arduino project that would be recognized as a multi-touch screen by Windows when I saw #123 . After messing around with information from there and a whole bunch of random places on the Internet, I am able to get it work on my PC by adding a new device under the frameworks of this library. Here's my code that does it and I hope you find it useful.

A few implementation notes not already mentioned in the comments:

- An abstract touchscreen has more parameters than something like a mouse/keyboard, notably max number of concurrent touches, but given the way these classes are instantiated and linked, I cannot figure out a simple way to let the user configure the parameters. It seems to me that user would inevitably have to edit library code, so for now I just placed two defines in MultiTouchAPI.h.
- For the Windows case, MSDN specifies that an HID Feature to provide "Maximum Contact Count" is mandatory, and it seems that without it Windows would assume the device as single-touch and ignore or give weird reactions when more than one touches is reported. So to implement HID Feature, I have to use the single report format and do USB stuff to respond feature requests. For single-touch digitizer the feature isn't necessary so a multi-report format may be used, but I didn't spare the effort to do it.
- The current descriptor reports 2 touches at a time, up to a total of 10 touches. It is identified in system properties as 10-point touchscreen. Each touch has coordinates, pressure, in-range flag. There are a few other optional usages that may also be included in the descriptor.

This is the first time I tried to work with USB, so please correct me if my understanding, code or anything is incorrect.

### DIFF
--- a/examples/MultiTouch/MultiTouch.ino
+++ b/examples/MultiTouch/MultiTouch.ino
@@ -1,0 +1,39 @@
+/*
+  Copyright (c) 2021 ilufang
+  See the readme for credit to other people.
+
+  Multi-touch touchscreen example
+  Draw 7 parallel lines across the screen.
+  Open Microsoft Whiteboard, select a brush and press the button.
+
+  See HID Project documentation for more infos
+  https://github.com/NicoHood/HID/wiki/Gamepad-API
+*/
+
+#include <HID-Project.h>
+
+const int pinButton = 2;
+
+void setup() {
+	pinMode(pinButton, INPUT_PULLUP);
+	MultiTouch.begin();
+}
+
+void loop() {
+	while(digitalRead(pinButton));
+
+	int16_t x = 2000, y = 2000;
+
+	for (; x <= 8000; x+=10) {
+		for (int i = 0; i < 7; i++) {
+			MultiTouch.setFinger(i, x, y+i*1000, (i+1)*15);
+		}
+		MultiTouch.send();
+		delay(10);
+	}
+
+	for (int i = 0; i < 7; i++) {
+		MultiTouch.releaseFinger(i);
+	}
+	MultiTouch.send();
+}
--- a/src/HID-APIs/MultiTouchAPI.h
+++ b/src/HID-APIs/MultiTouchAPI.h
@@ -1,0 +1,118 @@
+/*
+Copyright (c) 2021 ilufang
+See the readme for credit to other people.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+// Include guard
+#pragma once
+
+/// Maximum amount of fingers supported
+#define HID_MULTITOUCH_MAXFINGERS 10
+/// Number of fingers in a single report
+#define HID_MULTITOUCH_REPORTFINGERS 2
+
+// A report will always be the same size, even if you report fewer fingers than
+// REPORTFINGERS. The unused finger entries will simply be zero. If more fingers
+// are present than REPORTFINGERS, multiple reports will be sent to report all
+// fingers. This is know as "Hybrid Mode" on MSDN. The number of supported
+// fingers identified by Windows will still be MAXFINGERS. More than MAXFINGERS
+// contacts may be ignored by Windows even with hybrid mode.
+
+#define _MT_STATE_INACTIVE 0
+#define _MT_STATE_CONTACT  1
+#define _MT_STATE_RELEASED 2
+
+typedef struct ATTRIBUTE_PACKED {
+	uint8_t reportID;
+	uint8_t count;
+	struct ATTRIBUTE_PACKED {
+		uint8_t identifier;
+		uint8_t touch;
+		uint8_t pressure;
+		uint8_t x1, x0;
+		uint8_t y1, y0;
+	} contacts[HID_MULTITOUCH_REPORTFINGERS];
+} HID_MultiTouchReport_Data_t;
+
+typedef struct {
+	uint8_t status;
+	int8_t pressure;
+	int16_t x, y;
+} _finger_t;
+
+class MultiTouchAPI
+{
+public:
+	MultiTouchAPI() {
+		_fingers_count = 0;
+		for (int i = 0; i < HID_MULTITOUCH_MAXFINGERS; i++) {
+			_fingers[i].status = _MT_STATE_INACTIVE;
+		}
+	}
+
+	inline void begin();
+
+	/**
+	 * Set contact status for a finger in the internal data structure. You must
+	 * call send manually after setting all fingers to flush them through USB.
+	 *
+	 * @param id Finger id. Must be in the range of 0-MAXFINGERS. Same finger
+	 *           must have same id throughout contact. Allocations does not need
+	 *           to be continuous.
+	 * @param x, y Coordinates. Range 0-10000. (0,0) is top-left on Windows.
+	 * @param pressure Contact pressure. Range 0-127. When set to 0, the touch
+	 *                 is reported as hovering (in-range)
+	 * @return 1 if success. 0 if id is out-of-bounds
+	 */
+	inline int setFinger(uint8_t id, int16_t x, int16_t y, int8_t pressure=100);
+
+	/**
+	 * Release finger in the internal data structure. You must call send
+	 * manually after setting all fingers to flush them through USB.
+	 *
+	 * @param id Finger id. Must be in the range of 0-MAXFINGERS. Same finger
+	 *           must have same id throughout contact. Allocations does not need
+	 *           to be continuous.
+	 * @return 1 if success. 0 if id is out-of-bounds
+	 */
+	inline int releaseFinger(uint8_t id);
+
+	/**
+	 * Generates an HID report reflecting the currently recorded touch status
+	 * and send through USB.
+	 */
+	inline int send();
+
+protected:
+	/// Send the generated report. Needs to be implemented in a lower level
+	virtual int _sendReport() = 0;
+
+	/// Internal records of the current touch status
+	_finger_t _fingers[HID_MULTITOUCH_MAXFINGERS];
+	/// Number of active contacts, including just release contacts
+	uint8_t _fingers_count;
+
+	/// HID report to send. Not
+	HID_MultiTouchReport_Data_t _report;
+};
+
+// Implementation is inline
+#include "MultiTouchAPI.hpp"
--- a/src/HID-APIs/MultiTouchAPI.hpp
+++ b/src/HID-APIs/MultiTouchAPI.hpp
@@ -1,0 +1,105 @@
+/*
+Copyright (c) 2021 ilufang
+See the readme for credit to other people.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+// Include guard
+#pragma once
+
+#define _LSB(v) ((v >> 8) & 0xff)
+#define _MSB(v) (v & 0xff)
+
+void MultiTouchAPI::begin() {
+	send();
+}
+
+int MultiTouchAPI::setFinger(uint8_t id, int16_t x, int16_t y, int8_t pressure) {
+	if (id >= HID_MULTITOUCH_MAXFINGERS) {
+		return 0;
+	}
+	if (_fingers[id].status == _MT_STATE_INACTIVE) {
+		_fingers_count++;
+	}
+	_fingers[id].status = _MT_STATE_CONTACT;
+	_fingers[id].pressure = pressure;
+	_fingers[id].x = x;
+	_fingers[id].y = y;
+	return 1;
+}
+
+int MultiTouchAPI::releaseFinger(uint8_t id) {
+	if (id >= HID_MULTITOUCH_MAXFINGERS) {
+		return 0;
+	}
+	_fingers[id].status = _MT_STATE_RELEASED;
+	return 1;
+}
+
+int MultiTouchAPI::send() {
+	int ret = 0;
+
+	// Craft report(s)
+	_report.count = _fingers_count;
+
+	int rptentry=0;
+	for (int i = 0; i < HID_MULTITOUCH_MAXFINGERS; i++) {
+		if (_fingers[i].status == _MT_STATE_INACTIVE)
+			continue;
+
+		_report.contacts[rptentry].identifier = i; // valid for first report only
+
+		if (_fingers[i].status == _MT_STATE_RELEASED) {
+			// Released contacts need to be reported once with TipSW=0
+			_report.contacts[rptentry].touch = {};
+			_fingers_count--;
+			_fingers[i].status = _MT_STATE_INACTIVE;
+		} else {
+			// Active contacts must be reported even when not moved
+			_report.contacts[rptentry].touch = _fingers[i].pressure > 0 ? 3 : 1;
+			_report.contacts[rptentry].x1 = _MSB(_fingers[i].x);
+			_report.contacts[rptentry].x0 = _LSB(_fingers[i].x);
+			_report.contacts[rptentry].y1 = _MSB(_fingers[i].y);
+			_report.contacts[rptentry].y0 = _LSB(_fingers[i].y);
+			_report.contacts[rptentry].pressure = _fingers[i].pressure;
+		}
+
+		rptentry++;
+		if (rptentry == HID_MULTITOUCH_REPORTFINGERS) {
+			// Report full. Send now.
+			// If there are more contacts, they will be sent in subsequent
+			// reports with contact count set to 0
+			// See "Hybrid Mode" on MSDN docs
+			ret += _sendReport();
+			rptentry = 0;
+			_report.count = 0;
+		}
+	}
+
+	if (rptentry != 0) {
+		// Send remaining touches
+		for (; rptentry != HID_MULTITOUCH_REPORTFINGERS; rptentry++) {
+			_report.contacts[rptentry] = {};
+		}
+		ret += _sendReport();
+	}
+
+	return ret;
+}
--- a/src/SingleReport/MultiTouch.cpp
+++ b/src/SingleReport/MultiTouch.cpp
@@ -1,0 +1,246 @@
+/*
+Copyright (c) 2021 ilufang
+See the readme for credit to other people.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#include "MultiTouch.h"
+
+// HID Report identifiers
+#define _HID_REPORTID_TOUCH 0x01
+#define _HID_REPORTID_FEATURE 0x05
+
+// First part of the descriptor. It appears only once
+static const uint8_t _hidReportDescriptorTouchscreen_1[] PROGMEM = {
+	0x05, 0x0D,                    // USAGE_PAGE(Digitizers)
+	0x09, 0x04,                    // USAGE     (Touch Screen)
+	0xA1, 0x01,                    // COLLECTION(Application)
+
+	// define the actual amount of fingers that are concurrently touching the screen
+	0x85, _HID_REPORTID_TOUCH,     //   REPORT_ID (Touch)
+	0x09, 0x54,                    //   USAGE (Contact count)
+	0x25, 0x7f,                    //   LOGICAL_MAXIMUM (128)
+	0x95, 0x01,                    //   REPORT_COUNT(1)
+	0x75, 0x08,                    //   REPORT_SIZE (8)
+	0x81, 0x02                     //   INPUT (Data,Var,Abs)
+};
+
+// Finger definition part of the descriptor. It will repeat REPORTFINGERS times, once for each finger.
+// If more actual fingers are present than descriptor fingers, multiple reports will be sent sequentially under "Hybrid Mode" rule (See MSDN)
+static const uint8_t _hidReportDescriptorTouchscreen_2[] PROGMEM = {
+	// declare a finger collection
+	0x05, 0x0D,                    //   USAGE_PAGE(Digitizers)
+	0x09, 0x22,                    //   USAGE (Finger)
+	0xA1, 0x02,                    //   COLLECTION (Logical)
+
+	// declare an identifier for the finger
+	0x09, 0x51,                    //     USAGE (Contact Identifier)
+	0x25, 0x7f,                    //     LOGICAL_MAXIMUM (127)
+	0x75, 0x08,                    //     REPORT_SIZE (8)
+	0x95, 0x01,                    //     REPORT_COUNT (1)
+	0x81, 0x02,                    //     INPUT (Data,Var,Abs)
+
+	// declare Tip Switch & In range
+	0x09, 0x42,                    //     USAGE (Tip Switch)
+	0x09, 0x32,                    //     USAGE (In Range)
+	0x15, 0x00,                    //     LOGICAL_MINIMUM (0)
+	0x25, 0x01,                    //     LOGICAL_MAXIMUM (1)
+	0x75, 0x01,                    //     REPORT_SIZE (1)
+	0x95, 0x02,                    //     REPORT_COUNT(2)
+	0x81, 0x02,                    //     INPUT (Data,Var,Abs)
+	// declare the 6 padding bits as constant so the driver will ignore them
+	0x95, 0x06,                    //     REPORT_COUNT (6)
+	0x81, 0x03,                    //     INPUT (Cnst,Ary,Abs)
+
+	// declare pressure
+	0x09, 0x30,                    //     USAGE (Pressure)
+	0x25, 0x7f,                    //     LOGICAL_MAXIMUM (127)
+	0x75, 0x08,                    //     REPORT_SIZE (8)
+	0x95, 0x01,                    //     REPORT_COUNT (1)
+	0x81, 0x02,                    //     INPUT (Data,Var,Abs)
+
+	// define absolute X and Y coordinates of 16 bit each (percent values multiplied with 100)
+	0x05, 0x01,                    //     USAGE_PAGE (Generic Desktop)
+	0x09, 0x30,                    //     Usage (X)
+	0x09, 0x31,                    //     Usage (Y)
+	0x16, 0x00, 0x00,              //     Logical Minimum (0)
+	0x26, 0x10, 0x27,              //     Logical Maximum (10000)
+	0x36, 0x00, 0x00,              //     Physical Minimum (0)
+	0x46, 0x10, 0x27,              //     Physical Maximum (10000)
+	0x66, 0x00, 0x00,              //     UNIT (None)
+	0x75, 0x10,                    //     Report Size (16),
+	0x95, 0x02,                    //     Report Count (2),
+	0x81, 0x02,                    //     Input (Data,Var,Abs)
+	0xC0                           //   END_COLLECTION
+};
+
+// Last of the descriptor. It appears only once
+static const uint8_t _hidReportDescriptorTouchscreen_3[] PROGMEM = {
+	// define the maximum amount of fingers that the device supports
+	0x05, 0x0D,                    //   USAGE_PAGE(Digitizers)
+	0x85, _HID_REPORTID_FEATURE,   //   REPORT_ID (Feature)
+	0x09, 0x55,                    //   USAGE (Contact Count Maximum)
+	0x25, 0x7f,                    //   LOGICAL_MAXIMUM (127)
+	0xB1, 0x02,                    //   FEATURE (Data,Var,Abs)
+
+	0xC0                           // END_COLLECTION
+};
+
+
+MultiTouch_::MultiTouch_(void) : PluggableUSBModule(2, 1, epType), protocol(HID_REPORT_PROTOCOL), idle(1), featureReport((uint8_t *)&_ccmFeature), featureLength(sizeof(_ccmFeature))
+{
+	_report.reportID = _HID_REPORTID_TOUCH;
+	_ccmFeature.reportID = _HID_REPORTID_FEATURE;
+	_ccmFeature.contactCountMaximum = HID_MULTITOUCH_MAXFINGERS;
+	epType[0] = EP_TYPE_INTERRUPT_IN;
+	PluggableUSB().plug(this);
+}
+
+int MultiTouch_::getInterface(uint8_t* interfaceCount)
+{
+	*interfaceCount += 1; // uses 1
+	HIDDescriptor hidInterface = {
+		D_INTERFACE(pluggedInterface, 1, USB_DEVICE_CLASS_HUMAN_INTERFACE, HID_SUBCLASS_NONE, HID_PROTOCOL_NONE),
+		D_HIDREPORT(sizeof(_hidReportDescriptorTouchscreen_1) + sizeof(_hidReportDescriptorTouchscreen_2) * HID_MULTITOUCH_REPORTFINGERS + sizeof(_hidReportDescriptorTouchscreen_3)),
+		D_ENDPOINT(USB_ENDPOINT_IN(pluggedEndpoint), USB_ENDPOINT_TYPE_INTERRUPT, USB_EP_SIZE, 0x01)
+	};
+	return USB_SendControl(0, &hidInterface, sizeof(hidInterface));
+}
+
+int MultiTouch_::getDescriptor(USBSetup& setup)
+{
+	// Check if this is a HID Class Descriptor request
+	if (setup.bmRequestType != REQUEST_DEVICETOHOST_STANDARD_INTERFACE) { return 0; }
+	if (setup.wValueH != HID_REPORT_DESCRIPTOR_TYPE) { return 0; }
+
+	// In a HID Class Descriptor wIndex cointains the interface number
+	if (setup.wIndex != pluggedInterface) { return 0; }
+
+	// Reset the protocol on reenumeration. Normally the host should not assume the state of the protocol
+	// due to the USB specs, but Windows and Linux just assumes its in report mode.
+	protocol = HID_REPORT_PROTOCOL;
+
+	// Transmit HID descriptor. See comments next to the descriptor parts
+	int ret = 0;
+	ret += USB_SendControl(TRANSFER_PGM, _hidReportDescriptorTouchscreen_1, sizeof(_hidReportDescriptorTouchscreen_1));
+	for (int i = 0; i < HID_MULTITOUCH_REPORTFINGERS; i++) {
+		ret += USB_SendControl(TRANSFER_PGM, _hidReportDescriptorTouchscreen_2, sizeof(_hidReportDescriptorTouchscreen_2));
+	}
+	ret += USB_SendControl(TRANSFER_PGM, _hidReportDescriptorTouchscreen_3, sizeof(_hidReportDescriptorTouchscreen_3));
+
+	return ret;
+}
+
+bool MultiTouch_::setup(USBSetup& setup)
+{
+	if (pluggedInterface != setup.wIndex) {
+		return false;
+	}
+
+	uint8_t request = setup.bRequest;
+	uint8_t requestType = setup.bmRequestType;
+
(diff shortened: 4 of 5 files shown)


src/HID-APIs/MultiTouchAPI.h
```diff
@@ -0,0 +1,118 @@
...
+	 *           to be continuous.
+	 * @return 1 if success. 0 if id is out-of-bounds
+	 */
+	inline int releaseFinger(uint8_t id);
+
+	/**
+	 * Generates an HID report reflecting the currently recorded touch status
+	 * and send through USB.
+	 */
+	inline int send();
```
NicoHood, line 102: I am not sure if those API names can't be improved. Checkout the other APIs:
https://github.com/NicoHood/HID/blob/master/src/HID-APIs/KeyboardAPI.h#L48-L56
https://github.com/NicoHood/HID/blob/master/src/HID-APIs/AbsoluteMouseAPI.h#L69-L75

What about having:
```
// Move and press
press(uint8_t id, int16_t x, int16_t y, int8_t pressure=100);
press(uint8_t id, int8_t pressure=100);

// Move, press and release
tap(uint8_t id, int16_t x, int16_t y, int8_t pressure=100);
tap(uint8_t id, int8_t pressure=100);

// Release (after press)
release(uint8_t id);
releaseAll();
      
// Sending is public in the base class for advanced users.
virtual void SendReport(void* data, int length) = 0;
```

src/HID-APIs/MultiTouchAPI.h
```diff
@@ -0,0 +1,118 @@
...
+protected:
+	/// Send the generated report. Needs to be implemented in a lower level
+	virtual int _sendReport() = 0;
+
+	/// Internal records of the current touch status
+	_finger_t _fingers[HID_MULTITOUCH_MAXFINGERS];
+	/// Number of active contacts, including just release contacts
+	uint8_t _fingers_count;
+
+	/// HID report to send. Not
```
NicoHood, line 113: Why is there a "not"?

Also I think there is duplicated information here. The report contains the same information as the _fingers array. Those can be merged.

src/HID-APIs/MultiTouchAPI.hpp
```diff
@@ -0,0 +1,105 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+// Include guard
+#pragma once
+
+#define _LSB(v) ((v >> 8) & 0xff)
+#define _MSB(v) (v & 0xff)
```
NicoHood, line 28: Please inline those.

src/HID-APIs/MultiTouchAPI.h
```diff
@@ -0,0 +1,118 @@
...
+// A report will always be the same size, even if you report fewer fingers than
+// REPORTFINGERS. The unused finger entries will simply be zero. If more fingers
+// are present than REPORTFINGERS, multiple reports will be sent to report all
+// fingers. This is know as "Hybrid Mode" on MSDN. The number of supported
+// fingers identified by Windows will still be MAXFINGERS. More than MAXFINGERS
+// contacts may be ignored by Windows even with hybrid mode.
+
+#define _MT_STATE_INACTIVE 0
+#define _MT_STATE_CONTACT  1
+#define _MT_STATE_RELEASED 2
```
NicoHood, line 41: What about those? What is the difference between those states and are they used by the windows client? If yes, why dont we expose these?

Nevertheless, please put them in an enum.

src/HID-APIs/MultiTouchAPI.hpp
```diff
@@ -0,0 +1,105 @@
...
+		if (_fingers[i].status == _MT_STATE_RELEASED) {
+			// Released contacts need to be reported once with TipSW=0
+			_report.contacts[rptentry].touch = {};
+			_fingers_count--;
+			_fingers[i].status = _MT_STATE_INACTIVE;
+		} else {
+			// Active contacts must be reported even when not moved
+			_report.contacts[rptentry].touch = _fingers[i].pressure > 0 ? 3 : 1;
+			_report.contacts[rptentry].x1 = _MSB(_fingers[i].x);
+			_report.contacts[rptentry].x0 = _LSB(_fingers[i].x);
```
NicoHood, line 78: I think if you declare this x1 and x0 variable as int16, there is no need to to the MSB LSB shifting.

Like here:
https://github.com/NicoHood/HID/blob/master/src/HID-APIs/AbsoluteMouseAPI.h#L47-L48

src/SingleReport/MultiTouch.cpp
```diff
@@ -0,0 +1,246 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#include "MultiTouch.h"
+
+// HID Report identifiers
+#define _HID_REPORTID_TOUCH 0x01
```
NicoHood, line 27: This report ID is already used:
https://github.com/NicoHood/HID/blob/4bf6cd6167b1152a292bbc62793a85ab69070895/src/HID-Settings.h#L34-L36

What you are doing here is using a Multireport Device, not a single report it seems. Compare:
https://github.com/NicoHood/HID/blob/master/src/MultiReport/ImprovedMouse.cpp#L32
https://github.com/NicoHood/HID/blob/master/src/SingleReport/BootMouse.cpp#L31

So it would make sense to remove that line 37 from the descriptor. Please test, if it still works, I dont know that.

src/SingleReport/MultiTouch.cpp
```diff
@@ -0,0 +1,246 @@
...
+#define _HID_REPORTID_FEATURE 0x05
+
+// First part of the descriptor. It appears only once
+static const uint8_t _hidReportDescriptorTouchscreen_1[] PROGMEM = {
```
NicoHood, line 31: Suggested name: `_hidReportDescriptorMultiTouch`, to keep in sync with the other names. Or maybe you could rename multitouch to touchscreen? I guess that is a much better name and reflex the suggested API changed better (tap for example).

src/SingleReport/MultiTouch.cpp
```diff
@@ -0,0 +1,246 @@
...
+	0x85, _HID_REPORTID_FEATURE,   //   REPORT_ID (Feature)
+	0x09, 0x55,                    //   USAGE (Contact Count Maximum)
+	0x25, 0x7f,                    //   LOGICAL_MAXIMUM (127)
+	0xB1, 0x02,                    //   FEATURE (Data,Var,Abs)
+
+	0xC0                           // END_COLLECTION
+};
+
+
+MultiTouch_::MultiTouch_(void) : PluggableUSBModule(2, 1, epType), protocol(HID_REPORT_PROTOCOL), idle(1), featureReport((uint8_t *)&_ccmFeature), featureLength(sizeof(_ccmFeature))
```
NicoHood, line 107: Give me a quick hint what `2, 1` parameters mean.

src/SingleReport/MultiTouch.cpp
```diff
@@ -0,0 +1,246 @@
...
+{
+	_report.reportID = _HID_REPORTID_TOUCH;
```
NicoHood, line 109: This report ID is not required then. And even if it were, you can send it via:
https://github.com/NicoHood/HID/blob/master/src/MultiReport/AbsoluteMouse.cpp#L73-L76

src/SingleReport/MultiTouch.cpp
```diff
@@ -0,0 +1,246 @@
...
+	epType[0] = EP_TYPE_INTERRUPT_IN;
+	PluggableUSB().plug(this);
+}
+
+int MultiTouch_::getInterface(uint8_t* interfaceCount)
+{
+	*interfaceCount += 1; // uses 1
+	HIDDescriptor hidInterface = {
+		D_INTERFACE(pluggedInterface, 1, USB_DEVICE_CLASS_HUMAN_INTERFACE, HID_SUBCLASS_NONE, HID_PROTOCOL_NONE),
+		D_HIDREPORT(sizeof(_hidReportDescriptorTouchscreen_1) + sizeof(_hidReportDescriptorTouchscreen_2) * HID_MULTITOUCH_REPORTFINGERS + sizeof(_hidReportDescriptorTouchscreen_3)),
```
NicoHood, line 121: maybe split this up in 3 lines and add brackets for the multiplication?

src/SingleReport/MultiTouch.cpp
```diff
@@ -0,0 +1,246 @@
...
+		D_ENDPOINT(USB_ENDPOINT_IN(pluggedEndpoint), USB_ENDPOINT_TYPE_INTERRUPT, USB_EP_SIZE, 0x01)
+	};
+	return USB_SendControl(0, &hidInterface, sizeof(hidInterface));
```
NicoHood, line 124: What is this 0 used for again? Give me a hint please.

src/SingleReport/MultiTouch.cpp
```diff
@@ -0,0 +1,246 @@
...
+			return true;
+		}
+		if (request == HID_GET_IDLE) {
+			// TODO improve
+#ifdef __AVR__
+			UEDATX = idle;
+#endif
+			return true;
+		}
+	}
```
NicoHood, line 184: Now I have a question on that:
My implementation is lacking support for the different modes. I guess this is also the reason why some keyboards do not work in bios mode. Can you please tell me more about that and why is this special AVR case added? Maybe we can improve the keyboard and mouse API as well?
https://github.com/NicoHood/HID/blob/master/src/SingleReport/BootMouse.cpp#L99-L109

src/SingleReport/MultiTouch.cpp
```diff
@@ -0,0 +1,246 @@
...
+			// Feature (set feature report)
+			if(setup.wValueH == HID_REPORT_TYPE_FEATURE){
+				// No need to check for negative featureLength values,
+				// except the host tries to send more then 32k bytes.
+				// We dont have that much ram anyways.
+				if (length == featureLength) {
+					USB_RecvControl(featureReport, featureLength);
+					return true;
+				}
+				// TODO fake clear data?
```
NicoHood, line 210: What do you mean by that?

src/SingleReport/MultiTouch.cpp
```diff
@@ -0,0 +1,246 @@
...
+			// Output (not applicable)
+			else if(setup.wValueH == HID_REPORT_TYPE_OUTPUT){
+				return true;
+			}
+
+			// Input (set HID report)
+			else if(setup.wValueH == HID_REPORT_TYPE_INPUT)
+			{
+				if(length == sizeof(_report)){
+					USB_RecvControl(&_report, length);
```
NicoHood, line 222: Will the PC send back a report? I do not think this will ever happen, does it?

src/SingleReport/MultiTouch.h
```diff
@@ -0,0 +1,62 @@
...
+	// Implementation of the PUSBListNode
+	int getInterface(uint8_t* interfaceCount);
+	int getDescriptor(USBSetup& setup);
+	bool setup(USBSetup& setup);
+
+	EPTYPE_DESCRIPTOR_SIZE epType[1];
+	uint8_t protocol;
+	uint8_t idle;
+
+	uint8_t* featureReport;
```
NicoHood, line 52: What is this pointer used for? I am wondering if this points to unallocated space and will eventually cause undefined behavior. I might be wrong.

src/SingleReport/MultiTouch.cpp
```diff
@@ -0,0 +1,246 @@
...
+	return false;
+}
+
+uint8_t MultiTouch_::getProtocol() {
+	return protocol;
+}
+
+int MultiTouch_::_sendReport() {
+	return USB_Send(pluggedEndpoint | TRANSFER_RELEASE, &_report, sizeof(_report));
+}
```
NicoHood, line 238: I'd love to have a more generic:
https://github.com/NicoHood/HID/blob/master/src/MultiReport/AbsoluteMouse.cpp#L73-L76

This is what all other APIs in this library support.

NicoHood (changes requested):
Thanks a lot for submitting this very awesome patch!

I've taken a look at it and added lots of comments. I hope that helps in improving the code. Please keep in mind that I have not looked at my own code for years, so I might be wrong or missunderstand. You are the expert, as you've worked on that for the last weeks or even month.

A few major things after looking at this:
* What about renaming `MultiTouch` to `Touchscreen`? I guess that describes it better, right?
* Can you please tell me what this feature report is used for? Is it really required? I do not yet see the benefit for that, but I can imagine that Windows might need it?
* I've lot looked that close at the API implementation, as I'd suggest an API change that is more clear (see my comment). I don't know if my suggestion makes sense, but if, I will give it another review.

So lots of comments were made, I hope it was not too much. I am looking forward to get this integrated soon and as clean as possible. You made a really great job, USB is super tricky and overall this patch looks super good!

I am on vacation for multiple (2-4) weeks now, so I am sorry if I do not respond in time. Just wanted to let you know, as I really would like to continue on this after my vacation! If I miss the post, please ping me again.

Thanks a lot!

src/HID-APIs/MultiTouchAPI.h
```diff
@@ -0,0 +1,118 @@
...
+		uint8_t pressure;
+		uint8_t x1, x0;
+		uint8_t y1, y0;
+	} contacts[HID_MULTITOUCH_REPORTFINGERS];
+} HID_MultiTouchReport_Data_t;
+
+typedef struct {
+	uint8_t status;
+	int8_t pressure;
+	int16_t x, y;
```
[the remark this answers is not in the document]
ilufang: I was a bit unsure about endianness and signedness. I'll verify the details and clean this up

src/HID-APIs/MultiTouchAPI.h
```diff
@@ -0,0 +1,118 @@
...
+
+class MultiTouchAPI
+{
+public:
+	MultiTouchAPI() {
+		_fingers_count = 0;
+		for (int i = 0; i < HID_MULTITOUCH_MAXFINGERS; i++) {
+			_fingers[i].status = _MT_STATE_INACTIVE;
+		}
+	}
```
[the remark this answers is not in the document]
ilufang: Having the definition separately but in the same unit of translation causes multiple definition problems during linking. But since now all data components are zeros I'll just remove it altogether.

src/SingleReport/MultiTouch.cpp
ilufang: Sorry, typo. Should be `1,1`. It's just like everything else. They should be number of endpoints and interfaces.

src/SingleReport/MultiTouch.cpp
ilufang: The only nonzero value here I saw is `TRANSFER_PGM`, which means the 2nd parameter is a PROGMEM pointer. Otherwise this has been 0 everywhere.

src/SingleReport/MultiTouch.cpp
ilufang: Those were copied and I didn't touch it. Sorry I have no idea why it's there either.

src/SingleReport/MultiTouch.cpp
ilufang: It was from 
https://github.com/NicoHood/HID/blob/f2caf266c715fd42be14f0127822be3931e758de/src/SingleReport/BootKeyboard.cpp#L165 and I didn't clean up. I'll remove it.

src/SingleReport/MultiTouch.cpp
ilufang: I don't think so. It's just that BootKeyboard has it like this. Actually I don't think any of the Host2Device HID_SET_REPORT control requests make sense (except the input messages for BookKeyboard leds). I can remove them but I'm just not 100% sure that they are unnecessary.

src/SingleReport/MultiTouch.h
ilufang: It just points to the contactCountMaximum feature. I'll replace any usages with just direct pointer to that feature data.

src/HID-APIs/MultiTouchAPI.h
ilufang: 1. Typo
2. "Hybrid Mode" issue below

examples/MultiTouch/MultiTouch.ino
```diff
@@ -0,0 +1,39 @@
...
+	MultiTouch.begin();
+}
+
+void loop() {
+	while(digitalRead(pinButton));
+
+	int16_t x = 2000, y = 2000;
+
+	for (; x <= 8000; x+=10) {
+		for (int i = 0; i < 7; i++) {
```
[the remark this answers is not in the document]
ilufang: "Hybrid Mode" issue below.

2 fingers will fully exercise the report. Anything more than 2 exercises the Hybrid Mode reporting. I went over an issue with reporting not-multiple-of-2 fingers when unused fingers are not zeroed, so I decided to report 7 here. This will have fully populated reports, multiple reports for hybrid mode and partially populated reports.

src/SingleReport/MultiTouch.cpp
ilufang: I think if we have to do it like this for PluggableUSB.

src/SingleReport/MultiTouch.cpp
ilufang: See "Feature Report" issue below.

src/HID-APIs/MultiTouchAPI.h
ilufang: No they are just for internal tracking. 0 and 1 just says if an entry is in use or not. 2 helps to identify a "release" action because windows expects that a released touch is reported once with tip-switch set off off after releasing, instead of just being silently excluded from the report.

ilufang:
Thanks for the comments. I admit that the PR code is some quick implementation and not really polished. I'll get to the details soon. I'm not in any rush to get them merged but will do them sooner than later. Enjoy your vacation!

If you haven't worked with touch digitizers before I can imagine some of the details are not very straightforward compared to other HID devices, especially when Microsoft outlines additional requirements. The main two issues are "Feature Report" and "Hybrid Mode", and I'll summarize them here.

**Feature Report.** [Microsoft requires it](https://docs.microsoft.com/en-us/windows-hardware/design/component-guidelines/touchscreen-required-hid-top-level-collections). If you recall in the discussion of #123 Eric posted a working gist of 10 points for Android. It's my starting point, but Windows see it as a single touch device. I think the feature report is what Windows uses to decide a touchscreen is more than single-touch.

In order to support feature reports, I made this "single-report". It's not really for omitting report-id, but for PluggableUSB because PluggableHID does not seem to give me the opportunity to respond to a feature report request. I still included report-ids in the descriptor because that page also lists report-id as mandatory, and I'm not sure if feature reports would work without report-id. Honestly I was trying to implement feature reports by retrying random stuff relevant until my System Properties says 10-point support rather than single-touch support, and I'm not sure if I'm doing it actually correctly as the USB docs are too intimidating to navigate through.

**Hybrid Mode**. In my submitted code by default, each report only contains 2 touches, but a maximum 10 touches can be reported. When a single report does not contain all touches, hybrid mode allows sending multiple small reports in series to report all touches. Of course one could create a descriptor to contain all possible touches, but I was looking at a case of max ~40 touches but only ~5 is expected normally. Having a descriptor that big seems unnecessary and I'm not sure if there is some size limit on it. Details of how it works can be found [here](https://docs.microsoft.com/en-us/windows-hardware/design/component-guidelines/touchscreen-packet-reporting-modes).

With this reporting mode, one report does not contain all information. So I have a different data structure that tracks all contacts. The actual report(s) is generated from it during send. I think we can eliminate the report member storage from the class, but a H2D control transfer may set the input report and referenced it. It doesn't make sense to me why the host OS ever want to set a input message, and if we are sure a set input report won't happen we can remove it.

Let me know you have further issues, and I look forward to making it part of this awesome library!

Thanks.

src/SingleReport/MultiTouch.cpp
NicoHood: Alright, so this makes more sense then

NicoHood:
You can use a feature report within every hid device. For example I am using it optional in the keyboard:
https://github.com/NicoHood/HID/blob/master/src/SingleReport/BootKeyboard.cpp#L154

Now the question is, how to create a valid report, that contains the feature description in a single report. I think this was the tool, that can be used to build reports:
https://www.usb.org/document-library/hid-descriptor-tool
It only works for windows I think, but its worth testing out.

Nevertheless I think your solution could just work be removing the report IDs. But if not that should be also fine, as its an interface on its own.

---

If we change the API to my suggested solution (assuming my idea would work), we do not need to keep track of the pointer states, right? This way we can remove all structs to track the data. I'd leave that tracking up to the usercode, if possible. This saves us lots of ram. If that is our decision, we must of course remove the input callback. It see that you've copied it from the keyboard, I also dont know why it was there. I think for bootmode support, but that is not important here.

I think it would make most sense if you could cleanup this code first and then continue discussing :-)

src/SingleReport/MultiTouch.cpp
NicoHood: Oh I see, those are from the keyboard API. I think that must be removed with care. I suggest implementing it as the mouse, linked above.

examples/MultiTouch/MultiTouch.ino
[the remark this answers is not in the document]
NicoHood: And why do you define 10 fingers then? I think I dont understand that correct.

src/HID-APIs/MultiTouchAPI.h
NicoHood: If my suggest API implementation will work, I think this tracking is no longer required.

src/SingleReport/MultiTouch.cpp
NicoHood: I think you do not even need to receive that feature report at all, because it is never used, or is it??

src/SingleReport/MultiTouch.cpp
NicoHood: It is a special case for bootkeyboard. You can remove it. Just test and if it works, keep it like that.

src/SingleReport/MultiTouch.cpp
NicoHood: Oh I mean the function definition, not the content. Maybe more like this:
https://github.com/NicoHood/HID/blob/master/src/SingleReport/SingleAbsoluteMouse.cpp#L137-L140

examples/MultiTouch/MultiTouch.ino
```diff
@@ -0,0 +1,39 @@
+/*
+  Copyright (c) 2021 ilufang
+  See the readme for credit to other people.
+
+  Multi-touch touchscreen example
+  Draw 7 parallel lines across the screen.
+  Open Microsoft Whiteboard, select a brush and press the button.
+
+  See HID Project documentation for more infos
+  https://github.com/NicoHood/HID/wiki/Gamepad-API
```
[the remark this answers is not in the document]
ilufang: I just placed the wiki homepage there. If availability permits maybe we should put up a wiki page for this.

examples/MultiTouch/MultiTouch.ino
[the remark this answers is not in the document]
ilufang: In the current state, 2 is the number of fingers in the report, defined by the descriptor; 10 is the theoretical maximum number of fingers, defined by the feature report:

* Each report has 2 fingers
* If more than 2 fingers is present, we send more reports. Hybrid mode defines how to do it.
* We can send as many reports as desired, but Windows won't process more than 10 (software restriction I guess)
* Windows will say 10-finger support in System Properties
* We don't have a 10-finger report because it will be unnecessarily large for the common case where only 2 fingers are expected. Maybe there are also size limits so we can't pack like 100 touches into a single report, but still able to report them through 20 reports with 5 touches in each.

I'm setting 2/10 because I think that represents the most common scenarios. In fact I'm trying to do something that sends ~8 touches regularly up to max ~36 concurrent touches. I was expecting the user to configure these parameters.

src/HID-APIs/MultiTouchAPI.h
```diff
@@ -0,0 +1,118 @@
...
+#define _MT_STATE_RELEASED 2
+
+typedef struct ATTRIBUTE_PACKED {
+	uint8_t reportID;
+	uint8_t count;
+	struct ATTRIBUTE_PACKED {
+		uint8_t identifier;
+		uint8_t touch;
+		uint8_t pressure;
+		uint8_t x1, x0;
```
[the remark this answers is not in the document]
ilufang: Done

src/HID-APIs/MultiTouchAPI.h
```diff
@@ -0,0 +1,118 @@
...
+// are present than REPORTFINGERS, multiple reports will be sent to report all
+// fingers. This is know as "Hybrid Mode" on MSDN. The number of supported
+// fingers identified by Windows will still be MAXFINGERS. More than MAXFINGERS
+// contacts may be ignored by Windows even with hybrid mode.
+
+#define _MT_STATE_INACTIVE 0
+#define _MT_STATE_CONTACT  1
+#define _MT_STATE_RELEASED 2
+
+typedef struct ATTRIBUTE_PACKED {
```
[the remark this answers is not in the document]
ilufang: Done

src/HID-APIs/MultiTouchAPI.h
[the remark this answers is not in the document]
ilufang: Done

src/HID-APIs/MultiTouchAPI.h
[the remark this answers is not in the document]
ilufang: Constructor removed because default constructor will correctly initialize all data components.

src/HID-APIs/MultiTouchAPI.h
ilufang: I'm not yet very about having logic to perform sequences of operations to implement gestures. I'm sticking with the simplest operations for now. I'll come back to this when I find a chance.

src/HID-APIs/MultiTouchAPI.h
ilufang: The report contains less info than the _fingers array. See other thread about internal finger tracking

src/HID-APIs/MultiTouchAPI.hpp
ilufang: Removed

src/HID-APIs/MultiTouchAPI.hpp
ilufang: Done

src/SingleReport/MultiTouch.cpp
ilufang: It turns out that if I remove the report-ids of both input report and feature report, the HID device will still work. I have them deleted now.

But unfortunately I think we can't make a multi-report version due to the inability of responding to feature report requests under PluggableHID.

src/SingleReport/MultiTouch.cpp
ilufang: Everything else renamed to Touchscreen

src/SingleReport/MultiTouch.cpp
ilufang: Fixed.

src/SingleReport/MultiTouch.cpp
ilufang: All report-ids removed. See other thread about report-id

src/SingleReport/MultiTouch.cpp
ilufang: Done

src/SingleReport/MultiTouch.cpp
ilufang: Done

src/SingleReport/MultiTouch.cpp
ilufang: Yeah no reasonable scenario to receive any input or feature report here. I removed them.

src/SingleReport/MultiTouch.cpp
ilufang: Removed.

src/SingleReport/MultiTouch.h
ilufang: Done.

src/SingleReport/MultiTouch.cpp
ilufang: Done.

src/HID-APIs/MultiTouchAPI.h
ilufang: They are moved to hpp as an enum, but I think the current internal tracking cannot be further simplified. See below.

ilufang:
I've pushed my cleanups.

I'm not sure if I understood you correctly, but feature report is a concept parallel to input reports. A feature is not part of an input report. Feature reports travel through the control interface while input reports travel through the input interrupt interface (except when being polled which we don't handle and the OS normally don't do). You can see that it's part of setup and sent with `USB_SendControl` unlike the main input report.

It works with all report-ids removed, although MSDN specifies them as mandatory...

For the API I'm afraid it doesn't work nicely. An individual report contains less information than the fingers array due to hybrid mode. The send procedure is more like encoding the current state into a series of reports in a particular format and pattern. Although the reports can be parse back to touches, but it requires encoding and decoding and is not so straightforward like other HID descriptors. The input required to generate the reports is all active touches and just-released touches. Given the Arduino environment, first I don't think the user would be happy to provide a linked list or condensed array of active fingers. So an array of both active and inactive touches seems required. In that case, in terms of memory usage I don't think the user can possibly use less memory. (If they support less fingers, they should change the parameter) I believe offloading the tracking task to the user creates confusion, and I don't really see further simplifications to the current API.

Regarding higher-levels APIs to make gestures in a single call, I think it's definitely doable and can be relatively easily implemented with current APIs, but I'm unsure what would be a good collection of gestures. If the user needed most simple single-point gestures I think an AbsoluteMouse may be better. Multi-point gestures seems too diverse to cover in a few functions, especially that blocks when the gesture is being performed. If you have ideas feel free to just add them.

examples/MultiTouch/MultiTouch.ino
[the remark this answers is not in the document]
NicoHood: If you write a wiki page, I am happy to upload it. Just send me a markdown file.

src/HID-APIs/TouchscreenAPI.h
```diff
@@ -0,0 +1,115 @@
...
+	uint16_t y;
+} HID_Touchscreen_Finger_t;
+
+typedef union ATTRIBUTE_PACKED {
+	uint8_t whole8[0];
+	uint16_t whole16[0];
+	uint32_t whole32[0];
+	struct ATTRIBUTE_PACKED {
+		uint8_t count;
+		struct ATTRIBUTE_PACKED {
```
NicoHood, line 57: I dont think we need multiple attribute packes here. But please verify.

src/HID-APIs/TouchscreenAPI.hpp
```diff
@@ -0,0 +1,109 @@
...
+		if (_fingers[i].status == _MT_STATE_RELEASED) {
+			// Released contacts need to be reported once with TipSW=0
+			report.contacts[rptentry].touch = {};
+			_fingers_count--;
+			_fingers[i].status = _MT_STATE_INACTIVE;
+		} else {
+			// Active contacts must be reported even when not moved
+			report.contacts[rptentry].touch.status = HID_TOUCHSCREEN_TOUCH_IN_RANGE;
+			if (_fingers[i].pressure > 0)
+				report.contacts[rptentry].touch.status |= HID_TOUCHSCREEN_TOUCH_CONTACT;
```
NicoHood, line 82: Could you please add {} around?

src/HID-APIs/TouchscreenAPI.hpp
```diff
@@ -0,0 +1,109 @@
...
+	int ret = 0;
+	HID_TouchscreenReport_Data_t report;
+
+	// Craft report(s)
+	report.count = _fingers_count;
+
+	int rptentry=0;
+	for (int i = 0; i < HID_TOUCHSCREEN_MAXFINGERS; i++) {
+		if (_fingers[i].status == _MT_STATE_INACTIVE)
+			continue;
```
NicoHood, line 69: Could you please add {} around?

src/HID-Settings.h
```diff
@@ -73,6 +73,7 @@ THE SOFTWARE.
 #define HID_REPORTID_SURFACEDIAL 10
 #endif
 
+
```
NicoHood, line 76: Not required

src/HID-APIs/TouchscreenAPI.h
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * Generates an HID report reflecting the currently recorded touch status
+	 * and send through USB.
+	 */
+	inline int send();
+
+	/// Send generated report. Needs to be implemented in a lower level
+	virtual int sendReport(void *report, int length) = 0;
+	virtual int sendReport(HID_TouchscreenReport_Data_t &report) = 0;
```
NicoHood, line 103: This function is not required, as we already have the other. It gives an additional overhead, since it is virtual. Please remove it and use the other function.

src/HID-APIs/TouchscreenAPI.h
```diff
@@ -0,0 +1,115 @@
...
+		} contacts[HID_TOUCHSCREEN_REPORTFINGERS];
+	};
+} HID_TouchscreenReport_Data_t;
+
+class TouchscreenAPI
+{
+public:
+
+	inline void begin();
+
```
NicoHood, line 69: You have no end() function here, please add one. It should fully reset the touchscreen pressed.

src/HID-APIs/TouchscreenAPI.h
```diff
@@ -0,0 +1,115 @@
...
+		struct ATTRIBUTE_PACKED {
+			uint8_t identifier;
+			HID_Touchscreen_Finger_t touch;
+		} contacts[HID_TOUCHSCREEN_REPORTFINGERS];
+	};
+} HID_TouchscreenReport_Data_t;
+
+class TouchscreenAPI
+{
+public:
```
NicoHood, line 66: I've not written c++ for a while, but why dont you have a constructor? I can see, that I am clearing the data in the other constructors, why dont we do that here as well?
https://github.com/NicoHood/HID/blob/master/src/HID-APIs/MouseAPI.hpp#L27

src/SingleReport/Touchscreen.cpp
```diff
@@ -0,0 +1,204 @@
...
+	return protocol;
+}
+
+int Touchscreen_::sendReport(void *report, int length) {
+	return USB_Send(pluggedEndpoint | TRANSFER_RELEASE, report, length);
+}
+
+int Touchscreen_::sendReport(HID_TouchscreenReport_Data_t &report) {
+	return USB_Send(pluggedEndpoint | TRANSFER_RELEASE, &report, sizeof(HID_TouchscreenReport_Data_t));
+}
```
NicoHood, line 196: Can be removed

NicoHood:
> I've pushed my cleanups.

Good work!

> 
> I'm not sure if I understood you correctly, but feature report is a concept parallel to input reports. A feature is not part of an input report. Feature reports travel through the control interface while input reports travel through the input interrupt interface (except when being polled which we don't handle and the OS normally don't do). You can see that it's part of setup and sent with `USB_SendControl` unlike the main input report.

Correct!

> 
> It works with all report-ids removed, although MSDN specifies them as mandatory...

USB is a mystery! If it works, that is great. (Still confirms I have not forgotten everything about USB XD )

> 
> For the API I'm afraid it doesn't work nicely. An individual report contains less information than the fingers array due to hybrid mode. The send procedure is more like encoding the current state into a series of reports in a particular format and pattern. Although the reports can be parse back to touches, but it requires encoding and decoding and is not so straightforward like other HID descriptors. The input required to generate the reports is all active touches and just-released touches. Given the Arduino environment, first I don't think the user would be happy to provide a linked list or condensed array of active fingers. So an array of both active and inactive touches seems required. In that case, in terms of memory usage I don't think the user can possibly use less memory. (If they support less fingers, they should change the parameter) I believe offloading the tracking task to the user creates confusion, and I don't really see further simplifications to the current API.
> 
> Regarding higher-levels APIs to make gestures in a single call, I think it's definitely doable and can be relatively easily implemented with current APIs, but I'm unsure what would be a good collection of gestures. If the user needed most simple single-point gestures I think an AbsoluteMouse may be better. Multi-point gestures seems too diverse to cover in a few functions, especially that blocks when the gesture is being performed. If you have ideas feel free to just add them.

Now I am very new to this, and I really need to do deeper into this. If I could, I would really like to improve the API. But I am not sure if I am really right with that idea, it could be really that I dont understand the concept. The reason why I'd like to discuss this, is that I do not want to add a breaking change to the API later on, so we should design a good API now!

The rest of the PR should be fine now, I think. Just correct the few more annotations and we are mostly good to go.

NicoHood:
Please make sure to also add the touchscreen to the list of supported devices in the readme.

src/HID-APIs/TouchscreenAPI.h
ilufang: Did a quick google and [someone tried](https://stackoverflow.com/a/16904614) and found it gives different behaviors. Didn't test the behavior on Arduino but I'd stay on the safe side and keep it, especially when attribute_packed is not explicitly spelled out in C specs.

src/HID-APIs/TouchscreenAPI.h
ilufang: OK. Added.

src/HID-APIs/TouchscreenAPI.h
ilufang: Implicit default constructor is specified to zero everything in C++. But for consistency I added one anyway.

ilufang:
All fixes has been pushed.

I don't feel like designing a optimal high-level API at this moment because I don't really understand the high-level demands either. My use case has to directly use the setFinger and releaseFinger and they appear low-level enough that almost any higher-level APIs would call them. So I think we are good to publish with only these two essential APIs, and leave higher-level gestures open until there's a clearer pattern of uses or someone come up with a better design.

src/HID-APIs/TouchscreenAPI.h
NicoHood: I remember that this enlarges the sketches flash, is that still correct? I am wondering why I've added it in the first place then...

src/HID-APIs/TouchscreenAPI.h
NicoHood: Inline should be fine. You still want to keep it, hm? :-D

src/HID-APIs/TouchscreenAPI.h
ilufang: I shouldn't as it's the exact same behavior as the default constructor. I believe the compiler can figure out that static linking is all required for the global instances and no extra assign-to-zero code would be generated.

src/HID-APIs/TouchscreenAPI.h
ilufang: I think conceptually sendReport should take a report instead of an arbitrary buffer, but the buffer one is lower-level thus more suitable for being virtual. If inline is honored by the compiler, there should be no overhead of it being there.

src/HID-APIs/TouchscreenAPI.h
NicoHood: Can you test that real quick? It it adds additional overhead we should remove it from all other classes as well, I'd say.

ilufang: The explicit initializer list in the constructor (not the constructor itself) made the example program text 32 bytes larger. In disassembly gcc inserted runtime code to zero memory. If the commonsense in x86 that global data are zeroed during exe image loading also applies to AVR (in my test instance it does), then it is preferable to not include an explicit initializer list. 

I'm afraid I do not have the free time to perform project-wide polishing for you. I've removed the initializer list from this case. If there are no other touchscreen-specific issues, could you merge and publish this and then perform any other optimizations you'd like? Thanks.

NicoHood:
Could you please also integrate this patch:
https://github.com/NicoHood/HID/pull/310

ilufang:
> Could you please also integrate this patch:
> #310

Merged.

NicoHood:
What I meant was that you should also apply the fixes made to the other device, to your touchscreen device.

ilufang:
> What I meant was that you should also apply the fixes made to the other device, to your touchscreen device.

Done

sobrinho:
@NicoHood ping again :)

ilufang:
@tozz88 Looks like pluggableUSB functions are not being linked. IDE 1.8.13 should be fine. Are you sure you are using a board with native USB support? Do the rest of the HID library (especially single report devices) work fine?

NicoHood:
Hey guys,
I've seen your pings, but I am sorry, I ain't got time to review this again. And I do not want to merge it without having a closer review, I'd love to keep some quality.

tozz88:
> @tozz88 Looks like pluggableUSB functions are not being linked. IDE 1.8.13 should be fine. Are you sure you are using a board with native USB support? Do the rest of the HID library (especially single report devices) work fine?

@ilufang, Yes I'm using an MKRZERO board and have been using Multi-Report devices in this library successfully for years. It is SamD21 based. However, if I try any of the Single-Report devices I get the 4 missing entry points. It doesn't have to be Touchscreen. I tried SingleAbsoluteMouse and it also fails the same way. I assume Multi-Report also depends on pluiggableUSB? Is there some way I can inspect my pluggableUSB library? where does it reside on my system?

Is it possible that SamD21-based boards support Multi-Report but not Single-Report? 

thanks for your help,

ilufang:
@tozz88 I did a quick search through the Arduino cores and it appears that samd has a different interface than avr for some of its usb functions:

* PluggableHID, which MultiReport devices depends on, has a consistent interface
* PluggableUSB, which SingleReport devices and PluggableHID depend on, has a consistent but incomplete interface. To build a working device, core USB routine calls are required in addition to PluggableUSB methods.
* Core USB libraries are C++-style in SAMD but C-style in AVR and obviously differ. The [casually written documentation of PluggableUSB](https://github.com/arduino/Arduino/wiki/PluggableUSB-and-PluggableHID-howto#writing-a-library-as-a-pluggableusb-module) just used the AVR interfaces and I guess everyone read it assumed it to be standard.

So if you compare the [PluggableHID implementation of AVR](https://github.com/arduino/ArduinoCore-avr/blob/master/libraries/HID/src/HID.cpp) to [that of SAMD](https://github.com/arduino/ArduinoCore-samd/blob/master/libraries/HID/HID.cpp), you can see that a few calls to core usb routines may need to be replaced. Ideally Arduino library maintainers should make a better PluggableUSB, but for the time being, you can either edit the Touchscreen source, #define replace the USB_* functions, or if you prefer not to touch the library, link wrapper functions of the missing symbols. (BTW, don't you get warnings of using those functions without a header declaration? If so, make sure to fix the headers too) Unfortunately I do not have an SAMD-based device to test, so good luck on doing the workaround.

sobrinho:
@NicoHood any news?

Issue #225, opened 2020-07-01
KVM Switch


bitboy85:
I'm trying to use an Arduino micro as Bootkeyboard to control a KVM Switch. It seems the Keyboard is not recognized at all.
Are there different modes i could try?

The KVM support a Hotkey which is 2x CTRL_LEFT + 1 to 4 would be glad getting it to work so i can enter the same input on multiple pcs.

The code works fine on Windows and could be visualized with the on screen keyboard.

bitboy85:
After editing a few files die arduino keyboard is at least passed through the kvm and made available to the clients.
The KVM Hotkeys to switch the PC are still not working.
My newer Test machine doesn't show a keyboard error if the device is connected.

I compared a few settings/values with this tool: https://www.thesycon.de/eng/usb_descriptordumper.shtml between the micro and a standard usb keyboard.

Possible reasons
- The arduino doesn't enter "Bootmode" I've tried few changes to force bootmode but had no success
- The keycodes for those hotkeys are not correctly send (works in windows). Needed is Scroll_Lock and/or LEFT_CTRL
- The interface number is wrong. I got an interface number of 2, the real keyboard uses an interface id of 0 for boot mode. Sadly i couldn't find the place in the code where i could change it
- PAKET_SIZE is wrong. A micro uses by default 64, but a real keyboard only 8

What i've done so far. I don't know if all these steps are nessessary but i will document them for anyone who finds this by googling.

- I have removed CDC by applying this patch: https://github.com/arduino/Arduino/issues/6387#issuecomment-309363314
- removed bootloader by using another arduino as ISP
- applied patches provided here: https://github.com/NicoHood/HID/issues/102#issuecomment-508813744
- changed device descriptor. A real keyboard gives a device class of 0x00, a subclass of 0x00, and a protocol of 0x00. A default micro shows up different.
The required changes are in the USBCore.cpp
add 4 lines and change line 77:

```
#define DEVICE_CLASS      0x00 // Default 0xEF
#define DEVICE_SUB_CLASS  0x00 // Default 0x02
#define DEVICE_PROTOCOL   0x00 // Default 0x01
#define MAX_PAKET_SIZE    64 // Default 64

D_DEVICE(DEVICE_CLASS,DEVICE_SUB_CLASS,DEVICE_PROTOCOL,MAX_PAKET_SIZE,USB_VID,USB_PID,0x100,IMANUFACTURER,IPRODUCT,ISERIAL,1);
```

[...]

Issue #34, opened 2015-05-02
Keyboard not recognized at boot / BIOS


NicoHood:
The keyboard has to be a special boot keyboard with a standard descriptor. I heard it is possible with Teensy. More information can be found in the official USB docs.

```
http://www.usb.org/developers/hidpage/HID1_11.pdf
Seite 89 im PDF, 79 auf dem Blatt. Geht hier um Boot compatible HID devices.
Mit USB kann man ein Keyboard auf verschiedene Arten beschreiben. Standardmäßig 6 Modifier (strg, alt), 1 reserved und 6 keys. Das nutzt auch der Arduino. 

Auf seite 85/75 steht auch etwas feines. Das USB Gerät muss einen Boot Modus unterstützen, bzw sich als solches ausweisen.
Desweiteren steht hier, dass es mit dem Teensy wohl gehen sollte:
http://stackoverflow.com/questions/12682429/generic-keyboard-emulation-using-arduino-leonardo 
```

RingoM:
Hey Nico,

Have a look at Appendix E of the hid1_11 pdf you've linked to, there's an example:

>The sample device is a low-speed 105-key keyboard with an integrated pointing
>device. This device could be built using just one interface. However, two are used
>in this example so the device can support the boot protocol. As a result there are
>two Interface, Endpoint, HID and Report descriptors for this device.

NicoHood:
Sure, if I ever have time to do that. I didnt even had time to review your PR. I just add issues if things like that come to my mind. I also did not translate my german notes I wrote someone as email.

As always: If you have a fix, open a PR.

RingoM:
Haha, stupid 24 hour days ;-)

NicoHood:
I have to admit that I am working on the IRLremote instead :D totally worth it. Fixes will come, sooner or later.

bam80:
The feature would be very great!
Hope you will have time for this @NicoHood 
Thanks for your work

NicoHood:
Does the teensy work at boot? has anyone tested this? I dont know what is required for a boot keyboard. If the teensy works at boot we could try to search for the difference in the code and fix it.

The teensy keyboard is currently in the dev repo, but I dont think this works like this. I think the hid interface descriptor will be different then.

bam80:
In the Teensy USB Keyboard example it's clearly said:
"This code also supports the keyboard "boot protocol" ":
https://www.pjrc.com/teensy/usb_keyboard.html

[...]

PR #305, opened 2021-07-13
Improved compatibility with Mac

mdevaev:
Using HID-Project on Mac was associated with two very big problems. This patch solves both - for the boot mouse and the absolute mouse. I believe these fixes could be relevant for other devices, but I have only dealt with these two.

1. When using any HID-Project mouse in Recovery Mode, Mac does can see it at the kernel level (that is, the cursor moves), but at the same time it asks connecting the mouse (which is already connected). The fact is that checking for the connected mouse is performed in a very stupid way: by searching for `USAGE(Pointer)` in the descriptors. Hardware mouse always (or almost always) have this, so adding this to the descriptor solves the problem. Details [here](https://github.com/pikvm/pikvm/issues/289) and [here](https://community.twocanoes.com/t/mds-automaton-keyboard-mouse-bluetooth-screen-in-macos-11-3/6391/3). I added this for absolute and boot mouses and it generally improves compatibility. It probably makes sense to do this for all mouses.

2. Another problem is that Apple UEFI (it uses another driver) could not work with BootMouse at all because it did not see it. It turned out that he needed a response to the `HID_HID_DESCRIPTOR_TYPE` request (which is strange, since we are already passing this descriptor in the interface). This is really part of the HID standard, and hardware mouse respond to this. Usually the BIOS and other things do not require this, but Apple UEFI turned out to be finicky.

I also added a bit of protocol and idle processing. It is better to let it be than not to be.

This patch is the result of my work with the hardware USB protocol analyzer. To understand the subtleties of Apple drivers, I compared how the Arduino behaves and how a "live" mouse works. All this is tested, of course.

### DIFF
--- a/src/SingleReport/BootMouse.cpp
+++ b/src/SingleReport/BootMouse.cpp
@@ -29,6 +29,10 @@ static const uint8_t _hidReportDescriptorMouse[] PROGMEM = {
     0x09, 0x02,                      /* USAGE (Mouse) */
     0xa1, 0x01,                      /* COLLECTION (Application) */
 
+    /* Pointer and Physical are required by Apple Recovery */
+    0x09, 0x01,                      /*   USAGE (Pointer) */
+    0xa1, 0x00,                      /*   COLLECTION (Physical) */
+
 	/* 8 Buttons */
     0x05, 0x09,                      /*     USAGE_PAGE (Button) */
     0x19, 0x01,                      /*     USAGE_MINIMUM (Button 1) */
@@ -51,6 +55,7 @@ static const uint8_t _hidReportDescriptorMouse[] PROGMEM = {
     0x81, 0x06,                      /*     INPUT (Data,Var,Rel) */
 
 	/* End */
+    0xc0,                           /* END_COLLECTION (Physical) */
     0xc0                            /* END_COLLECTION */
 };
 
@@ -73,18 +78,24 @@ int BootMouse_::getInterface(uint8_t* interfaceCount)
 
 int BootMouse_::getDescriptor(USBSetup& setup)
 {
-	// Check if this is a HID Class Descriptor request
-	if (setup.bmRequestType != REQUEST_DEVICETOHOST_STANDARD_INTERFACE) { return 0; }
-	if (setup.wValueH != HID_REPORT_DESCRIPTOR_TYPE) { return 0; }
-
 	// In a HID Class Descriptor wIndex cointains the interface number
 	if (setup.wIndex != pluggedInterface) { return 0; }
 
-	// Reset the protocol on reenumeration. Normally the host should not assume the state of the protocol
-	// due to the USB specs, but Windows and Linux just assumes its in report mode.
-	protocol = HID_REPORT_PROTOCOL;
+	// Check if this is a HID Class Descriptor request
+	if (setup.bmRequestType != REQUEST_DEVICETOHOST_STANDARD_INTERFACE) { return 0; }
 
-	return USB_SendControl(TRANSFER_PGM, _hidReportDescriptorMouse, sizeof(_hidReportDescriptorMouse));
+	if (setup.wValueH == HID_HID_DESCRIPTOR_TYPE) {
+		// Apple UEFI wants it
+		HIDDescDescriptor desc = D_HIDREPORT(sizeof(_hidReportDescriptorMouse));
+		return USB_SendControl(0, &desc, sizeof(desc));
+	} else if (setup.wValueH == HID_REPORT_DESCRIPTOR_TYPE) {
+		// Reset the protocol on reenumeration. Normally the host should not assume the state of the protocol
+		// due to the USB specs, but Windows and Linux just assumes its in report mode.
+		protocol = HID_REPORT_PROTOCOL;
+		return USB_SendControl(TRANSFER_PGM, _hidReportDescriptorMouse, sizeof(_hidReportDescriptorMouse));
+	}
+
+	return 0;
 }
 
 bool BootMouse_::setup(USBSetup& setup)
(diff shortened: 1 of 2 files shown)


mdevaev:
Sorry for my terrible English, I'm very sleepy while writing this issue :)

src/SingleReport/BootMouse.cpp
```diff
@@ -73,18 +78,24 @@ int BootMouse_::getInterface(uint8_t* interfaceCount)
...
-	if (setup.wValueH != HID_REPORT_DESCRIPTOR_TYPE) { return 0; }
-
 	// In a HID Class Descriptor wIndex cointains the interface number
 	if (setup.wIndex != pluggedInterface) { return 0; }
 
-	// Reset the protocol on reenumeration. Normally the host should not assume the state of the protocol
-	// due to the USB specs, but Windows and Linux just assumes its in report mode.
-	protocol = HID_REPORT_PROTOCOL;
+	// Check if this is a HID Class Descriptor request
+	if (setup.bmRequestType != REQUEST_DEVICETOHOST_STANDARD_INTERFACE) { return 0; }
```
NicoHood, line 85: Why did you move this down?

src/SingleReport/BootMouse.cpp
```diff
@@ -73,18 +78,24 @@ int BootMouse_::getInterface(uint8_t* interfaceCount)
...
+
+	if (setup.wValueH == HID_HID_DESCRIPTOR_TYPE) {
+		// Apple UEFI wants it
+		HIDDescDescriptor desc = D_HIDREPORT(sizeof(_hidReportDescriptorMouse));
+		return USB_SendControl(0, &desc, sizeof(desc));
```
NicoHood, line 90: Do we need this for the bootkeyboard as well?

NicoHood:
Thanks a lot for sending this in!

It looks like you've got very good equipment and knowledge about USB (as much knowledge as you can have with that very weird spec + try &error). Would you mind having a look at these issue with the bootkeyboard? Maybe you have also a device that does not work, and the knowledge to fix that as well?

https://github.com/NicoHood/HID/issues/34
https://github.com/NicoHood/HID/issues/225

src/SingleReport/BootMouse.cpp
mdevaev: To make if-else branch for `setup.wValueH`. Previously, there was an inequality and a return, and now there is a check for two types of descriptor.

src/SingleReport/BootMouse.cpp
mdevaev: Totally yes. This is the second problem that I described in the issue. This is how hardware mice behave, I also found it in the datasheet for some [mouse chip](https://www.silabs.com/documents/public/application-notes/AN249.pdf). Almost all the devices I have personally met do not require this command, but all Macs do. It's funny that this constant exists in the HID library, but is not used. Apparently, no one thought that the command associated with it could be mandatory.

mdevaev:
To solve these problems, I need samples of problematic devices. Everything works perfectly on all the motherboards I have. I am aware of the problems with KVM, but not with the BIOS. Your library is used by many thousands of people through my project, and none of them said that there are problems with the BIOS and keyboard in the Arduino. On the contrary, I recommend Arduino for crappy BIOSes if the Linux Gadget does not work well.

I'll keep you posted. I'm even interested in finding out what's wrong there myself. If I get my hands on a broken device, I will investigate it and let you know. If you need anything, you can write to me by email mdevaev@gmail.com or discord https://discord.gg/bpmXfz5

NicoHood:
What is the project you are working on?

Did you notice my comments to the code?

mdevaev:
I'm doing [this](https://github.com/pikvm/pikvm), and [hardware for it](https://www.youtube.com/watch?v=dTchVKxx7Fo).

I responded to your comments in the review. Don't you see them?

src/SingleReport/BootMouse.cpp
NicoHood: Yes I saw this, but this check can be kept above as well, right? I mean it should not matter and I also do not prefer anything. It is just for my understanding, that this line could also be placed above the wIndex check, right?

src/SingleReport/BootMouse.cpp
NicoHood: So I guess you will be trying to fix the keyboard in another PR? I think this PR is ready to merge then, I dont see anything critical that should not work.

src/SingleReport/BootMouse.cpp
mdevaev: Yes, there is no difference. I put the index above for reasons that first you need to check whether there was a request to this device, and then proceed to check the type of request. It would be strange to check the type first, then the index, and then the other type.

src/SingleReport/BootMouse.cpp
mdevaev: Oh, crap, I'm sorry, I missed that you asked about the keyboard, and continued about the mouse. Stupid me. No, apparently this is not necessary for the keyboard. At least Apple and everything else works. But, POTENTIALLY, this may be due to that bug with KVM and strange BIOSes. Maybe if you add this code, everything will work. In the analysis of USB data, I have not seen a reference to this command, and I have not yet been able to find a dataset for some keyboard chip where this would be described. I would ask someone to experiment, but it's definitely not worth including this for the keyboard until we're sure.

Bugerman21:
Hi All,
One year ago I have used this library on Leonardo, to simulate keystrokes on the keyboard from windows, boot menu, and BIOS, and check Caps Lock led status.
Now, I flow a similar problem with a new BIOS on Dell laptops.
When I am entering the boot menu or the BIOS, the laptop did not get from Leonardo a keystrokes.

Maybe this will help to someone for investigating the problem.
**At this moment I found this solution that works with UNO:**
I am sending commands from Leonardo to UNO.
https://www.instructables.com/How-to-Make-a-Arduino-HID-Keyboard/

NicoHood:
This would possibly be the better solution:
https://github.com/NicoHood/HID/issues/225

PR #306, opened 2021-07-14
Windows 7 can't handle negative absolute values

mdevaev:
Just subj. When using negative values, only a quarter of the screen is available. All other OS and BIOSes work fine, but Windows 7 has a broken driver. The solution is taken from a drawing tablet - counting from zero. In addition, I have made sure that your API is backward compatible so that people can still use negative values.

In theory, this patch reduces the positioning accuracy, but 64K Super-Mega-Full-HD screens are unlikely to appear in the next 10 years for this to become a problem. Counting from zero works fine for all devices I know.

### DIFF
--- a/src/HID-APIs/AbsoluteMouseAPI.hpp
+++ b/src/HID-APIs/AbsoluteMouseAPI.hpp
@@ -78,8 +78,8 @@ void AbsoluteMouseAPI::moveTo(int x, int y, signed char wheel){
 	yAxis = y;
 	HID_MouseAbsoluteReport_Data_t report;
 	report.buttons = _buttons;
-	report.xAxis = x;
-	report.yAxis = y;
+	report.xAxis = ((long)x + 32768) / 2;
+	report.yAxis = ((long)y + 32768) / 2;
 	report.wheel = wheel;
 	SendReport(&report, sizeof(report));
 }
--- a/src/HID-APIs/AbsoluteMouseAPI.hpp
+++ b/src/HID-APIs/AbsoluteMouseAPI.hpp
@@ -80,7 +80,7 @@ void AbsoluteMouseAPI::moveTo(int x, int y, signed char wheel){
 	report.buttons = _buttons;
 	// The range -32768...32767 is converted to 0...32767 because Windows 7
 	// does not support negative coordinates.
-	// Negative values are supported here for compatibility reasons.
+	// Negative values are supported here for API compatibility reasons to keep (0,0) as a center of screen.
 	// See detauls in AbsoluteMouse sources and here: https://github.com/NicoHood/HID/pull/306
 	report.xAxis = ((int32_t)x + 32768) / 2;
 	report.yAxis = ((int32_t)y + 32768) / 2;
--- a/src/HID-APIs/AbsoluteMouseAPI.hpp
+++ b/src/HID-APIs/AbsoluteMouseAPI.hpp
@@ -80,6 +80,7 @@ void AbsoluteMouseAPI::moveTo(int x, int y, signed char wheel){
 	report.buttons = _buttons;
 	// The range -32768...32767 is converted to 0...32767 because Windows 7
 	// does not support negative coordinates.
+	// Negative values are supported here for compatibility reasons.
 	// See detauls in AbsoluteMouse sources and here: https://github.com/NicoHood/HID/pull/306
 	report.xAxis = ((int32_t)x + 32768) / 2;
 	report.yAxis = ((int32_t)y + 32768) / 2;
--- a/src/HID-APIs/AbsoluteMouseAPI.hpp
+++ b/src/HID-APIs/AbsoluteMouseAPI.hpp
@@ -78,8 +78,11 @@ void AbsoluteMouseAPI::moveTo(int x, int y, signed char wheel){
 	yAxis = y;
 	HID_MouseAbsoluteReport_Data_t report;
 	report.buttons = _buttons;
-	report.xAxis = ((long)x + 32768) / 2;
-	report.yAxis = ((long)y + 32768) / 2;
+	// The range -32768...32767 is converted to 0...32767 because Windows 7
+	// does not support negative coordinates.
+	// See detauls in AbsoluteMouse sources and here: https://github.com/NicoHood/HID/pull/306
+	report.xAxis = ((int32_t)x + 32768) / 2;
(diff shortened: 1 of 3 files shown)


src/HID-APIs/AbsoluteMouseAPI.hpp
```diff
@@ -78,8 +78,8 @@ void AbsoluteMouseAPI::moveTo(int x, int y, signed char wheel){
 	yAxis = y;
 	HID_MouseAbsoluteReport_Data_t report;
 	report.buttons = _buttons;
-	report.xAxis = x;
-	report.yAxis = y;
+	report.xAxis = ((long)x + 32768) / 2;
```
NicoHood, line 81: instead of long, please use uint32_t. I am also not use, if the number after should be castes as well. It would make sense to use a 0x?? like representation to understand it better. Please also add a comment that we are converting signed range (xx -yy) to unsigned range 0-0xFFFF

NicoHood:
Good catch! I just trust you in that case, as I dont have any windows PC. If the API is still compatible, that should be fine.

You wrote, that this patch reduced accuracy. Why?

src/HID-APIs/AbsoluteMouseAPI.hpp
mdevaev: > instead of long, please use uint32_t

Okay, but int32_t, when calculating, we still need a sign.

> 0x

Just for uniformity to avoid "stairs".

Also comment added.

NicoHood:
Whz dont you use 0-0xffff (65k) as new range for windows 7? This would result in the exact same resolution?

mdevaev:
Well, imagine, you have a range -32768...32767. That is, you can position a 65536x65536 screen with pixel accuracy. Now you are limited to 32768x32768. If you suddenly use HID somewhere in NASA or the distant future, where the resolution of your screen will be more than 32k pixels, then when you change the position by one pixel, the cursor will move by two :D

mdevaev:
> Why dont you use 0-0xffff (65k) as new range for windows 7? This would result in the exact same resolution?

I tried it, it doesn't work at all. Apparently, they have a two-byte limit.

NicoHood:
but 65k is 2 bytes!? I guess you wrote the descriptor just wrong?

src/HID-APIs/AbsoluteMouseAPI.hpp
```diff
@@ -80,6 +80,7 @@ void AbsoluteMouseAPI::moveTo(int x, int y, signed char wheel){
 	report.buttons = _buttons;
 	// The range -32768...32767 is converted to 0...32767 because Windows 7
 	// does not support negative coordinates.
+	// Negative values are supported here for compatibility reasons.
```
NicoHood, line 83: Well it is not only compatibility, it is the API. I wanted an API that works with 0,0 being in the center. "Nobody" cares about the underlying usb part, so this comment is not correct. I have no better idea how to describe it though

mdevaev:
I mean, apparently there is an upper limit in the driver in the form of two bytes, but it is signed. This is strange, because they do not support negative values. If you try to use 0xFFFF, then nothing will work. In my case, the mouse just didn't move. So I just set 0 for the lower bound and left 32767 for the upper one.

src/HID-APIs/AbsoluteMouseAPI.hpp
mdevaev: I tried to improve it

NicoHood:
What if you set the upper bound to lets say 40k? (something between 32 and 65)? What if you set the max to 0xFFFFE (-1)? Maybe just the max value is an issue?

Just WTF

src/HID-APIs/AbsoluteMouseAPI.hpp
```diff
@@ -80,7 +80,7 @@ void AbsoluteMouseAPI::moveTo(int x, int y, signed char wheel){
 	report.buttons = _buttons;
 	// The range -32768...32767 is converted to 0...32767 because Windows 7
 	// does not support negative coordinates.
-	// Negative values are supported here for compatibility reasons.
+	// Negative values are supported here for API compatibility reasons to keep (0,0) as a center of screen.
```
NicoHood, line 83: perfect!

mdevaev:
Presumably this should work, but most likely poorly written drivers will ruin everything. Unfortunately, I can't check it now. I haven't used Windows 7 for a long time. This patch that I'm offering is a fix that I made more than a year ago in my project. Since then, it has been used by all my users, so you can be sure that it is correct.
* https://github.com/pikvm/kvmd/blob/master/hid/patches/hid-abs-mouse-win-fix.patch
* https://github.com/pikvm/kvmd/blob/master/hid/src/main.cpp#L211

NicoHood:
Could you get a user that can test that? I'd like to test that.

mdevaev:
Check the patch or your assumption about 40k?

NicoHood:
The 40k, 0xFFFE assumption.

mdevaev:
I have one such user, but he is sleeping now. I'll be able to ask for it in a few hours.

But in any case, I would not change the value of 32767 to something else simply for the reasons that this is a guaranteed working value for all many systems, since it has been tested by many users. Even if 40k works, I don't think changing the upper bound would be a good idea. Who knows where it might break.

NicoHood:
Sure 40k is super stupid. I just want to know if there is just an issue with using the total maximum value 0xffff. But I guess you got what I mean ;-)

mdevaev:
Yea) Okay, if there are no more problems with this patch, could you merge it and make a new version? When we test the other values, I will let you know.

NicoHood:
Yeah, good idea. I dont expect that to change much, but would be happy about any new results.

mdevaev:
Make it also here please: https://platformio.org/lib/show/523/HID-Project

NicoHood:
I have nothing to do with platform.io. That is not my account.

mdevaev:
How interesting. It turns out that someone puts it out, but it is misleading, because it does not say who, but it says that you are the author. I thought it was yours. Okay.

NicoHood:
I think it will be pulled automatically every night or so. It seems you cannot even register for that.

mdevaev:
@NicoHood boop. We've checked it. Not working at all. Windows sees the mouse but does not process events from it. Apparently 32767 is an upper limit.

NicoHood:
Thanks for that report :-)

PR #310, opened 2021-07-23
Bugfix: SET_IDLE uses wValueH for duration, not wValueL

mdevaev:
Hello again :)

I found out that the duration in the SET_IDLE request is actually responsible for the highest byte, not the lowest. Here is an excerpt from the standard:
image

image

To verify this, I used [USBCV](https://www.usb.org/document-library/usb3cvx64-tool), the official tool for testing compliance with USB standards. Without this patch, the "Idle Test" was a failure, everything works as it should with the patch.

PS: Soon I will bring you a whole pack of new fixes to comply with the HID standard.

### DIFF
--- a/src/SingleReport/BootKeyboard.cpp
+++ b/src/SingleReport/BootKeyboard.cpp
@@ -142,7 +142,7 @@ bool BootKeyboard_::setup(USBSetup& setup)
 			return true;
 		}
 		if (request == HID_SET_IDLE) {
-			idle = setup.wValueL;
+			idle = setup.wValueH;
 			return true;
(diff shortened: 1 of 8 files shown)


NicoHood:
Thanks for reporting. I think this also needs to get fixed in the upstream HID core (AVR and all other MCU variants such as ARM as well):
https://github.com/arduino/ArduinoCore-avr/blob/master/libraries/HID/src/HID.cpp#L131

Very excited for all the other fixes!

mdevaev:
Yep, good idea. I sent PR.

PR #311, opened 2021-07-23
BootKeyboard, SingleAbsoluteMouse: handle HID_HID_DESCRIPTOR_TYPE

mdevaev:
Continuation of the story in #305.

Handling of `HID_HID_DESCRIPTOR_TYPE` is required by [USBCV](https://www.usb.org/document-library/usb3cvx64-tool) (the official USB tool for testing for compliance with the standard).

So yes, this is the answer to your question: it looks like processing this is required for all HID devices. I added for those two devices that I had the opportunity to test.

### DIFF
--- a/src/SingleReport/BootKeyboard.cpp
+++ b/src/SingleReport/BootKeyboard.cpp
@@ -90,18 +90,24 @@ int BootKeyboard_::getInterface(uint8_t* interfaceCount)
 
 int BootKeyboard_::getDescriptor(USBSetup& setup)
 {
-	// Check if this is a HID Class Descriptor request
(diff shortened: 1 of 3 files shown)


NicoHood:
Nice! Could you also have a look at the other Devices?

mdevaev:
I can, but it's unlikely to be soon. I mean, not earlier than in a month - I'm very busy with launching a startup. Sorry :(

But. Given that this is part of the standard, I think that you can add processing of this request to all HID without looking. In fact, it seems to me that there is a refactoring here for some general function that can be inserted into all the other HID.

NicoHood:
Alright, I will merge this and track the remaining devices in https://github.com/NicoHood/HID/issues/312

PR #314, opened 2021-08-14
Fixed absolute mouse positioning on Windows 98

mdevaev:
Sup! Another funny bug.

Windows 98 contains a crappy HID driver. It allows you to move only
on the upper right quarter of the screen. Yes, this is indeed exceeding
the maximum value from the HID report. Yes, it really should work that way.

VirtualBox has a similar fix, but with a shift in the other direction
(I didn't dig the details). I suppose it can be fixed somehow with a special
HID descriptor, but the proposed fix is simpler and really works.

Related: https://github.com/pikvm/pikvm/issues/159

### DIFF
--- a/src/HID-APIs/AbsoluteMouseAPI.h
+++ b/src/HID-APIs/AbsoluteMouseAPI.h
@@ -57,6 +57,7 @@ protected:
 	int16_t xAxis;
 	int16_t yAxis;
 	uint8_t _buttons;
+	bool win98_fix;
 	inline void buttons(uint8_t b);
 
 	inline int16_t qadd16(int16_t base, int16_t increment);
@@ -65,6 +66,8 @@ public:
 	inline AbsoluteMouseAPI(void);
 	inline void begin(void);
 	inline void end(void);
+	inline void setWin98FixEnabled(bool enabled);
+	inline bool isWin98FixEnabled(void);
 
 	inline void click(uint8_t b = MOUSE_LEFT);
 	inline void moveTo(int x, int y, signed char wheel = 0);
--- a/src/HID-APIs/AbsoluteMouseAPI.hpp
+++ b/src/HID-APIs/AbsoluteMouseAPI.hpp
@@ -51,7 +51,7 @@ int16_t AbsoluteMouseAPI::qadd16(int16_t base, int16_t increment) {
 }
 
 AbsoluteMouseAPI::AbsoluteMouseAPI(void):
-xAxis(0), yAxis(0), _buttons(0)
+xAxis(0), yAxis(0), _buttons(0), win98_fix(false)
 {
 	// Empty
 }
@@ -64,6 +64,14 @@ void AbsoluteMouseAPI::begin(void){
 void AbsoluteMouseAPI::end(void){
 	_buttons = 0;
 	moveTo(xAxis, yAxis, 0);
+}
+
+void AbsoluteMouseAPI::setWin98FixEnabled(bool enabled){
+	win98_fix = enabled;
+}
+
+bool AbsoluteMouseAPI::isWin98FixEnabled(void){
+	return win98_fix;
 }
 
 void AbsoluteMouseAPI::click(uint8_t b){
@@ -84,6 +92,13 @@ void AbsoluteMouseAPI::moveTo(int x, int y, signed char wheel){
 	// See detauls in AbsoluteMouse sources and here: https://github.com/NicoHood/HID/pull/306
 	report.xAxis = ((int32_t)x + 32768) / 2;
 	report.yAxis = ((int32_t)y + 32768) / 2;
+	if (win98_fix) {
+		// Windows 98 contains a buggy driver.
+		// Yes, this is indeed exceeding the maximum value from the HID report.
+		// Yes, it really should work that way.
+		report.xAxis <<= 1;
+		report.yAxis <<= 1;
+	}
 	report.wheel = wheel;
 	SendReport(&report, sizeof(report));
 }


mdevaev:
PS: It seems that it cannot be solved in a common way and I would not risk changing the handle for the sake of a dead OS and jeopardizing compatibility with modern systems.

xcasxcursex:
Is it just me or is this adding a comparison operation to every click event? (`if (win98_fix) `)

Perhaps this feature should be wrapped in defines and included at build time so that the 2 people in the universe with a legit reason to use win98 can get it if they need it but it doesn't impact performance for everyone else?

mdevaev:
@xcasxcursex Data transfer takes disproportionately longer than a single comparison operation. If you open the arduino core code, you will see that there is no point in saving ops on such things at all. Conditional compilation for such a thing is like trying to save two cents when buying a Lamborghini.

In addition, conditional compilation reduces the flexibility of the firmware. You can configure dynamic parameters from the firmware itself if you connect the arduino to different hosts, otherwise you will have to recompile and flash it every time.

xcasxcursex:
> @xcasxcursex Data transfer takes disproportionately longer than a single comparison operation.

I'm aware, but the magnitude of the wastage is not the issue here, the issue is wastage, at all. Of course, it's not wasted, _if_ you are running win98, but for everybody else, it's wastage.

> If you open the arduino core code, you will see that there is no point in saving ops on such things at all. 

There is a point, it's the performance of practically everything using the library. Saving ops everywhere is important - and this isn't even a matter of "saving" ops, this is "not wasting ops" - we're not talking about an optimisation of existing code here.

If you feel that the arduino core is wasting cycles, that is a reason to optimise the arduino core, not an excuse to waste more cycles elsewhere. 

> Conditional compilation for such a thing is like trying to save two cents when buying a Lamborghini.

No it's like trying to save two _seconds_ when _racing_ a lamborghini. The buy price would be a one-time cost. This is not like that at all. 

It is clear that performance is not your _primary_ concern here, but supporting an OS that's been dead for almost two decades is.... and that's fine, you run whatever OS you like and it's awesome that you can make it possible for others.... but I'm sure you realise this is not universal, or even vaguely common. This is more like slowing every car on the road so that you can ride a horse and buggy there safely.


Ask the users of this library the direct question: Would you prefer better performance, or win98 support? Those are the options here, and I think we both know what the answer will be - even if we replaced "better" with "marginally and almost immeasurably small amount better".

But that doesn't mean there's *no* value in win98 support of course. Conditional compilation at least offers the option to both parties and everyone can be satisfied. I'm not saying conditional compilation is the right or only way to solve this problem, but I am saying that negatively impacting everyone's performance, by however small amount it may be, for the sake of supporting an OS that shouldn't even in use any more, that is an actual problem, that needs consideration.... So... Consider it?  I'm sure if you're smart enough to make this patch you're smart enough to find a way to do it without a net negative result. I'm happy to help :)

mdevaev:
@xcasxcursex I don't think that saving on ONE operation from a couple of clock cycles is worth doing conditional compilation for it. You have an 8 or 16 mhz MCU, an arduino core with a giant overhead, and C++ firmware. This is not a PIC8! And are you trying to optimize a SINGLE if? In the operation that the user initiates? Neither the user nor the scripts will be able to generate significant HID events so quickly that it is generally noticeable even on a large scale. Are you trying to transfer terabytes of data via HID using morse code?

Even the code in this function, which provides compatibility with windows 7, takes more MCU clock cycles. Why not cut it out as well? After all, Microsoft stops supporting older OSS.

I'm not angry, it's just absurd.

xcasxcursex:
I understand why you feel it's absurd. It's a tiny thing, I get it. You don't need to try to drum that home.

What I don't understand is why you seem to want to avoid the fact that every SINGLE instruction adds up. This attitude of "oh it's only a little performance loss, so it doesn't matter at all", if you adopt it as a rule, is how to eventually end up with generally poorly performing code. No, it's only a little performance loss, so it only matters a little. But it DOES matter.

What I don't understand is how you won't consider that basically everybody (since basically nobody uses win98) would prefer not to have the extra instruction there at all.

What I find absurd, is that you don't even want to discuss alternatives that could keep everyone happy, and seem adamant that a net loss is the only way forward.


> Even the code in this function, which provides compatibility with windows 7, takes more MCU clock cycles. Why not cut it out as well?

Sure, why not? Or at least, consider it. I'm just here commenting on this change, for now. I might have made similar comments if I'd seen that change go through. I haven't looked at that, so I can't say. But that's another topic for another time. I'm here talking about this one, today.

> In the operation that the user initiates?

It's a mouse click operation, let's not imply that the user initiating it is anything other than expected, and expected to occur frequently, and expected to be as timely as possible.

xcasxcursex:
>  an arduino core with a giant overhead

And please, can we not do the thing where I listen to you and answer you and then you ignore me and repeat yourself? We did this already.

> If you feel that the arduino core is wasting cycles, that is a reason to optimise the arduino core, not an excuse to waste more cycles elsewhere.

Come on mate you're clearly far smarter than to believe that "it's already bad so it doesn't matter if we make it worse".

mdevaev:
Do you have an accurate idea of the duration of the comparison operation and how USB works? Did you know that USB HID is a polling protocol? What concept of "timely" can we talk about if the host polls your mouse at best every 1 millisecond? Your mouse on the table is most likely polled every 8 milliseconds. A single flag check statement will not have an impact that you could detect.

> What I find absurd, is that you don't even want to discuss alternatives that could keep everyone happy, and seem adamant that a net loss is the only way forward.

These are changes of the times of changes and premature optimization in its pure form. If you don't agree with me, then you can hardly argue with Donald Knuth:

> The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming.

You're trying to fix something that isn't broken, and you haven't convinced me. You say that I am smart enough, but at the same time you immediately push me to a solution that you like, and I do not agree with the need for it (due to experience, some knowledge about the USB protocol and understanding how the arduino core works). If you don't like my arguments - okay, no problem, you are entitled to your opinion.

Ultimately, the decision will be left to @NicoHood. If he says that saving a single IF makes sense and he wants it, I will add it to the patch.

JenniferHatches:
@xcasxcursex fwiw, I'd rather have the compatibility. You made some bets as to what people would prefer, so I'm going to give you my two cents:
Are we really arguing performance hit from adding a single if statement to code that runs on mouse clicks (user initiated)?  Because that would be *absurd*.  Please describe under exactly what scenario you would see any measurable performance loss. 

Surely, the *best* thing for a HID library is *compatibility*?  So it just works no matter what USB device you plug it into? Like, universally? I wonder if we could incorporate that into the name...

xcasxcursex:
Just a thought I had, might help to get the ball rolling on a solution to this:

I've talked above about one possible solution being conditional compilation, to include this . You mentioned another performance impact (win7 support) and It's plausible there are more (especially if the "it's only a small problem so we'll ignore it" thing has been the approach in the past as it was here). Would you be more inclined toward conditional compilation which would include these kinds of things by default, but provide a higher-performance option for those who _really_ care? 

Y'know, as in, not #IFDEF WIN98_FIX and #IFDEF WIN7_FIX and friends, or #IFDEF QUIRKS_OBSOLETE_OS, but #IFDEF PERFORMANCE_MODE

Of course I'm over here trying to discuss it and you're typing out a reply refusing to even discuss it. Are you being so obtuse intentionally?


> You're trying to fix something that isn't broken

No, I'm not. You just completely fail to see how it's broken, in spite of my clearly outlining it. This is why you're quoting Knuth, talking about time wasting. 

You think it's a waste because YOU don't think it's important, but you fail to see how the opinions of others, and the objective measure of it, don't agree. Every wasted cycle is wasted, opinions be damned. You speak of not having a detectable impact but you refuse to even consider others' use-cases so how could you assume the impact upon them? At least start to consider them. How much time have you wasted arguing when you could at least be working toward something preferable? 

Besides, if you carry Knuth's quote there to its logical conclusion, it is not to avoid the optimisation entirely, as you are, but to do it LATER ("at the wrong time" and "premature" imply this and the rest of his work explicitly states it). If you want to start an issue to do an optimisation pass as a way around the problems introduced by this change (and the win7 one and whatever else) then that might be the way forward. I don't know. Would be easier if you'd at least discuss it.

>  If he says that saving a single IF makes sense and he wants it, I will add it to the patch.

But otherwise you won't even discuss it because you can't or won't (I don't know) consider the prominent use-case of preferring performance over obsolete OS support. I maintain that you strike me as smarter than this, so it seems like you're actively trying to be difficult, dude. Why? Just talk it over, that's all I've asked of you.

> you immediately push me to a solution that you like,

That never happened, and I take exception to your suggesting it. I've specifically gone out of my way to make it very clear that I'm open to other alternatives. In fact as you can see I was just typing one up, because while you're sitting there arguing, I'm actually trying to make the code better, and thinking about how it might be done in a way that you'd like it. It's only you, who are adamant that their way is the only way. What alternatives have you suggested?

> experience, some knowledge about the USB protocol and understanding how the arduino core works

You're attempting to imply I lack these things because you don't agree with me, and you're wrong. This is the THIRD TIME you've brought up the shortcoming of the core and I'll reply the same way as I did the last times:

>If you feel that the arduino core is wasting cycles, that is a reason to optimise the arduino core, not an excuse to waste more cycles elsewhere.

With a reminder I kindly asked you not to do that.


Mate, do you need to take a while to relax and come back at this? I feel like you've maybe had a bad day and it's effecting your judgement and personal skills. I'm just trying to get you to talk it over.



> fwiw, I'd rather have the compatibility.

Same! If you think I'm arguing against the inclusion of the compatibility code, you've definitely misunderstood. I'm all for it.

> Surely, the _best_ thing for a HID library is _compatibility_? So it just works no matter what USB device you plug it into? Like, universally?

Well, that depends on who you ask. For a guy like me who is a sweaty nerd and expects millisecond accuracy and wants to have the device running an RTOS where this might impact the performance of motor drivers that need timely behaviour, and will literally never ever plug it into a win98 machine ever, no, I'd rather the performance, Personally. But that's just me (and most people). Let's not ignore the very real use-case of needing win98 support (what about someone making a cabinet sim who want to run a game that ONLY runs on 98? He matters too!)

Why not BOTH compatibility AND performance, or at least, the option to choose? I think choice of outcome is preferable to a fixed outcome. I mean imagine if the situation were reversed and you had ONLY the performance, and poor compatibility? That'd suck, right? Of course, which is why this change exists and is mostly good. It just needs some attention given to its very real problem.

JenniferHatches:
So your argument is that the (approximately 3-9 cycles worth of assembly instructions, depending) has the possibility of affecting the real time use on a device that runs at 8Mhz?

Interesting.

Also: if you'd ever worked in industry, like oil and gas, machining, other industrial equipment? Yeah, LOTS of it runs Win98.

I still have lots of equipment I admin that runs it. Not by choice.

mnotgninnep:
I'm the target audience here. I maintain Win98 and up machines. I also have linux devices and servers. A switchable option to control compatibility would be ideal. Having to maintain multiple images to then re-image or swap sd cards in the control device when switching between systems would be a complete pain. Unless there's a heavy performance cost, I'd go for compatibility here.

xcasxcursex:
> So your argument is that the (approximately 3-9 cycles worth of assembly instructions, depending) has the possibility of affecting the real time use on a device that runs at 8Mhz?

No, my argument is that sacrificing performance without even considering and discussing alternatives is not wise, that the minor performance impacts of myriad changes combine to amount to a major performance impact because if you never consider performance there's no reason why would it be an organic result, etc. I think I've made it pretty clear above, no?

Perhaps you've been confused by my giving a single example use-case (note I gave one of each?) and think that's the entire argument? Read the entire argument, I did put it all there. 

> Interesting.

Being condescending does not bolster your argument in any way, on the contrary, it gives the impression that since you're out of tech talk you've decided to try and make it personal.


> I posit that if you're actually looking at microsecond level timing, you'd probably want to write in assembly, given that you can't guarantee how the compiler will optimize...so please, just stop.

I posit that if you're actually looking at microsecond timing, and you don't actually look at it, and you do that a thousand times, now it's a millisecond. Keep it up, and that time grows and grows. How is it that people smart enough to cut code can't wrap their head around 'every drop in the ocean counts'? It's almost like your emotions are clouding your judgement.

I mean sure, if actually considering the performance impact had some exceeding cost, avoid it. But it doesn't. A few minutes to chat about it would do. You guys have spent WAY more time arguing with me than it would have taken to work with me and find a workable solution. 


> Also: if you'd ever worked in industry, like oil and gas, machining, other industrial equipment? Yeah, LOTS of it runs Win98.
> 
> I still have lots of equipment I admin that runs it. Not by choice.

Yeh I've seen it a ton in banks, my most recent run-in was an oil tanker. This isn't my first rodeo. I find it telling, that you guys are repeatedly attempting to attack my experience, and make strawman arguments like this one implying that I don't see the value of win98 support when I've clearly stated otherwise multiple times, rather than actually discuss the points I'm bringing up.

xcasxcursex:
> A switchable option to control compatibility would be ideal. Having to maintain multiple images to then re-image or swap sd cards in the control device when switching between systems would be a complete pain. Unless there's a heavy performance cost, I'd go for compatibility here.

Right, this is why my knee-jerk reaction to this was conditional compilation. For those who want compatibility you don't want to have to do something annoying, just build it that way and expect it to work everywhere. For those who know they're never going to see win98 again, they can build it that way and expect no cycles to be wasted on something they will never use.

Because there seems an argumentative vibe here I'll say it AGAIN: I' m not saying conditional compilation is the only way or best way or whatever, just that some solution should be considered and this is the first one to come to mind.

mdevaev:
@xcasxcursex There are no half measures if you want to get rid of the IF. Because the code is either compiled or not, and you are pushing to make it optional - that is, to agree with you.

Let's do a little experiment. We take an oscilloscope and calculate the execution time of each operation.
<details>
  <summary>CLICK ME TO EXPAND</summary>

1. In the main cycle, we raise the pin and lower it, we get a pulse duration of 125ns (0, 000 000 125 sec), at a frequency of 16 MHz it takes TWO cycles.
image

2. The main loop in the my firmware is processed in 6.5 microseconds (0, 000 006 520 sec) or 104 cycles.
image

3. The flag check wrapped in raising and lowering the pin (125ns) takes 690ns (0, 000 000 690 sec) or 9 cycles.
image

4. The time spent in the interrupt is 152.4us (0, 000 152 sec) or 2432 cycles, so adding a flag check increases the time spent in the interrupt by 0.37%
image

5. The interrupt repetition time is 100 ms (0.1 sec)
image

**All values are given with an error of +-10ns.**

</details>

**Conclusions:**
1. **Adding one IF increases the time spent in the interrupt by 0.37%**
2. **The IF execution time is ~584ns (0, 000 000 565sec)**

As I said this does not have any noticeable effect on performance in the case of HID code. The HID does not transmit data more often than 1ms by design. Compare this value and the time spent on IF.

I will not discuss it with you in this way anymore.

JenniferHatches:
> I posit that if you're actually looking at microsecond timing, and you don't actually look at it, and you do that a thousand times, now it's a millisecond. Keep it up, and that time grows and grows. How is it that people smart enough to cut code can't wrap their head around 'every drop in the ocean counts'? It's almost like your emotions are clouding your judgement.
>
No.  We literally are looking at it.  And oh look, you haven't countered any of my points.  You've never even come up with a single scenario where this has any impact.  You argue the "what ifs" of someone worrying about micro or nanosecond latency on a *human initiated action* (which hardware tests show you're wrong about).  In a device with millisecond polling...

So your argument is that when a human clicks a mouse at a precise time (humans are definitely microsecond accurate!), and the bus is polled 8 milliseconds later, that the demonstrated actual hardware time cost of this is relevant?


If not,*why are you talking?*


You're in denial, and are protecting all over the place. 

 I mean sure, if actually considering the performance impact had some exceeding cost, avoid it. But it doesn't. A few minutes to chat about it would do. You guys have spent WAY more time arguing with me than it would have taken to work with me and find a workable solution.
> 
> > Also: if you'd ever worked in industry, like oil and gas, machining, other industrial equipment? Yeah, LOTS of it runs Win98.
> > I still have lots of equipment I admin that runs it. Not by choice.
> 
> Yeh I've seen it a ton in banks, my most recent run-in was an oil tanker. This isn't my first rodeo. I find it telling, that you guys are repeatedly attempting to attack my experience, and make strawman arguments like this one implying that I don't see the value of win98 support when I've clearly stated otherwise multiple times, rather than actually discuss the points I'm bringing up.

xcasxcursex:
> How is it that people smart enough to cut code can't wrap their head around 'every drop in the ocean counts'? It's almost like your emotions are clouding your judgement.

You're kinda proving my point here.

Look, I;ve got better things to do than argue and I'm sure you do too. I'll leave with simply this:

You're performing a host operating system check on every mouse click. 

If you think that's the best way you could do it.............Go ahead.

JenniferHatches:
> > How is it that people smart enough to cut code can't wrap their head around 'every drop in the ocean counts'? It's almost like your emotions are clouding your judgement.
> 
> You're kinda proving my point here.
> 
> Look, I;ve got better things to do than argue and I'm sure you do too. I'll leave with simply this:
> 
> You're performing a host operating system check on every mouse click.
> 
> If you think that's the best way you could do it.............Go ahead.

Because *context matters*.

Every drop in the ocean doesn't matter. If I take one drop on particular out, *the ocean is not empty*.  Your analogies don't make sense.

xcasxcursex:
> Look, I;ve got better things to do than argue and I'm sure you do too. I'll leave with simply this:
> 
> You're performing a host operating system check on every mouse click.
> 
> If you think that's the best way you could do it.............Go ahead.

NicoHood:
Guys, I did not read your chat, I just checked the code. From my guess, this should take 1 more bytes of ram and eventually a few more bytes on flash (it is inline, so the compiler might remove it anyways, but I guess it will never remove the ram usage).

--> Can someone verify that please?

It seems that there is no automatic win98 detection, so the whole thing is forced by code anyways. So I don't see any reason why we could not use a `#define` here, other than letting the firmware switch to a windows mode dynamically at runtime when you press a button or so.

So lets assume we want that behavior: What if we just pass another bool to the click function? That would not take up more ram and also no flash, since the compiler will optimize the inline call. And even if it does we could do a `clickwin98fix()` function, that does a similar job. Then the firmware would be able to still support win98, but all other users are not affected of this thing.

So what are the options? (Sorry if you need to repeat, as I've not read the whole story above)

mdevaev:
@NicoHood The bottom line is that this fix does not concern the click, but the basic function of moving the mouse. In order not to increase the number of additional arguments in each function, I just added one class property that allows you to enable or disable win98 compatibility mode. This costs us only one IF (bool). There is no loss of performance here, and this is the most minimal and easiest way to support win95. All this discussion from above was with a random person who decided to save 565 damn nanoseconds.

mdevaev:
UPD @NicoHood 

> So I don't see any reason why we could not use a #define here, other than letting the firmware switch to a windows mode dynamically at runtime when you press a button or so.

This is a main and only reason. The implementation that I propose allows for firmware to enable compatibility mode dynamically, does not change the behavior of your existing API and does not break compatibility with existing code. The price is a single IF(bool) :)

xcasxcursex:
>   to save 565 damn nanoseconds.

This is literally the first line of my first reply to you.

>  the magnitude of the wastage is not the issue here, the issue is wastage, at all.



All I was trying to do here, is to actually consider the possibility of improving this code. 

Most of the above is explaining why that's a good idea (because a policy of accumulating wasted CPU cycles, is a path to poorly performing software - because accumulation works like that.  )

I'm doing my best to avoid argument here, but blatantly misquoting me is not helping. Want to stop with that? Since I've asked you to what, three times now?

mdevaev:
@xcasxcursex Just stop. There is no way to improve this code while preserving its functions. The code is either executed or not. To prevent the code from being executed, it must be enclosed in an ifdef. You are not calling for finding ways to improve the code, you are manipulating facts and playing with words to get people to agree with you.

**Once again: there are no performance issues here. Generally. At all.**

xcasxcursex:
> @xcasxcursex Just stop

I already did, but you insisted on an argument for some reason. You directly tell lies about me and then tell me to stop when I correct them. You're a joke.

> You are not calling for finding ways to improve the code,

Except for the literally dozens of times I have called for exactly that, and the fact that is literally all I ever asked for.

> you are manipulating facts and playing with words to get people to agree with you.

Given the blatant misquote you made up there, that statement sounds hilarious.

mdevaev:
@NicoHood 
> So lets assume we want that behavior: What if we just pass another bool to the click function? That would not take up more ram and also no flash, since the compiler will optimize the inline call. And even if it does we could do a clickwin98fix() function, that does a similar job. Then the firmware would be able to still support win98, but all other users are not affected of this thing.

You will have to duplicate all the APIs associated with click and move, and it will be inconvenient for the user code to work with this-a double set of functions. So you split this logic: part will be inside the HID library (bit shift and new API) and part (function selection) will go to the firmware codi. Code duplication and a new API where it is not required at all. And then, imagine that I will bring you some other fix, for example, for Novell NetwWare. Each new fix should not cause duplication of the entire API, it should be only one parameter.

@xcasxcursex I don't communicate with you. Go practice trolling somewhere else.

xcasxcursex:
> Go practice trolling somewhere else.

Yeh because I'm trolling when I correct your repeated blatant misquotes? Yeh dude, no, that's not what trolling means.

I'll give you a tip: You can keep talking BS about me and I'll keep correcting you, or you can just stop and I won't have any reason to be here.

mdevaev:
@NicoHood Sorry, since my explanations were again drowned in food, I will briefly summarize:

1. The price of the current solution is 565 nanoseconds per user action. HID cannot transmit data more often than once every 1ms by design, so there is no point in this saving. The user will never be able to enter data so quickly that it causes a noticeable delay.
2. Yes, this increases the memory consumption by a couple of dozen bytes. Attempts to optimize memory consumption do not make sense here, because if you want to optimize it, you need to start with the fatter parts of the arduino core and other things in the library.
3. Creating a new separate API specifically for win98 will be inconvenient for the user, because you will assign the selection logic to him (he will have to write if-else everywhere) and will require you to almost completely duplicate the click and move functions, but with the win98 prefix. The solution with property does not require code duplication, does not change your API and does not break compatibility. And it's just convenient - I use this function exactly like this in my code. I just set the flag, and all the firmware code works with win98 in the correct way.
4. Above, I gave the measurement results that show that this patch does not introduce any visible delay, and does not load the MCU even on my rather heavy firmware.

xcasxcursex:
> or you can just stop and I won't have any reason to be here.

> since my explanations were again drowned in food,

.... Try again.

NicoHood:
I totally missed, that `moveTo()` is being called multiple times.

I do not bother about any time delay, but I bother a minimal bit about the 1 byte of extra ram. Now I would like to support such a fix for the users, but **if** there is a better approach, we should go with it. So lets think of it again:

I understand that adding parameters is not fitting here. Usinge `#define` would work, but not at runtime. Is there a requirement for that?

I think there is possibly a solution, if we use c++ templates. We could define a special Win98AbsoluteMouseAPI that would support windows 98. Did you thought about this?

mdevaev:
@NicoHood If you create an entire object for the win95 case, you will significantly worsen memory consumption for me and those users who need it. Because with a high probability they will want to use both mice: both normal and for win98. This will lead to the duplication of a disproportionately large number of elements, to a significantly higher consumption of flash memory, and again will require a pile of if-else for code that wants to work with this API. A single bool variable is the cheapest solution that you can come up with. And this is not an abstract theory. I use this code for dynamic switching in my own firmware (quite heavy) and I still haven't come close to having to cut out memory for one variable at a time.

I measured the memory consumption with and without this feature. This is my own firmware so it is quite heavy and includes some additional code. But from the point of view of RAM, there is no difference at all based on the platformio report. I think this is due to the peculiarities of packaging structures in C++. Just look: even on my [Pi-KVM firmware](https://github.com/pikvm/kvmd/tree/master/hid/src) (which contains the code for working with SPI and other things), almost 10 copies of runtime will fit into the MCU memory before one byte becomes a problem.

**With this code**
```
RAM:   [=         ]  10.4% (used 267 bytes from 2560 bytes)
Flash: [====      ]  36.6% (used 10488 bytes from 28672 bytes)
```

**Without this code**
```
RAM:   [=         ]  10.4% (used 267 bytes from 2560 bytes)
Flash: [====      ]  36.9% (used 10578 bytes from 28672 bytes)
```

NicoHood:
Nice! I was not aware, that this additional ram byte is being compiled away. But anyways, you are right in general. Just wanted to bring that up, if it would be an option for you.

Just curios, why do you need windows 98 support?

mdevaev:
One of the users was here, he uses it for retro gaming. In addition, there are still a lot of industrial equipment that still runs on Windows 98 (at best lol)

mdevaev:
@NicoHood Could you make a new release with it? I need it for my public puild.

mdevaev:
@NicoHood boop

NicoHood:
Sorry, I am still hestitating to merge this, as I find the usecase so small and the fix quite bloaty.

Can't we fix that outside the library in a way moveTo() is called?

Looking more close at the moveTo operation:
1. It adds 32768 in order to convert it to a signed value.
2. It divides by 2 to get half resolution, as this is written in the USB report
3. Win98 fix: The left shift is actually multiplying again by 2

So a mathematical function would be:
```
Normal:
report = (y + 32768) / 2

Win98:
report = (x + 32768) / 2 * 2

We can transform the win98 function to:
report = (x + 32768) * 2 / 2 
report = (x*2 + 32768 *2) / 2 
report = (x*2 + 32768 + 32768)  / 2 
report = ((x*2 + 32768) + 32768)  / 2

Which will remind us of the original function:
report = (y + 32768) / 2
and y now becomes:
y = x*2 + 32768
```

So a possible solution would be (not tested of course):

```cpp
void AbsoluteMouseAPI::moveToWin98(int x, int y, signed char wheel){
  moveTo(x*2 + 32768, y*2 + 32768, wheel); // TODO: Add some typecasting to not exceed maximums
}
```

This wrapper is way smaller and more beautiful. In the c# world (if you are familar with c#) this could be an extension method for the class.

mdevaev:
If you do this, you will force the client code to choose between two functions. Moreover, higher-level functions such as click and move will use the usual moveTo. Duplicating the API is a very bad idea, since you will force the user to duplicate his code and call different functions, instead of processing one flag in a single place. I also demonstrated that adding a parameter has no effect on memory consumption.

But I have another idea. I've optimized the code a bit. Look again.

NicoHood:
The click function should still work, as the MoveTo() will also save this wrong function.

Since the user code must set/unset win98 fix I see no difference in calling a different function from outside. You need some if/else logic in the usercode anyways. But it will not bloat the library.

mdevaev:
At a minimum, you will have to duplicate the `move` function. If you make an additional method (or several) then you will still add something to the library. But from the user's point of view, the best solution is to minimize the API and avoid duplicating it. I did this based on how it would be more convenient to use it. It really doesn't seem to me that fixing a few lines is a bloat.

This approach with API duplication is bad because you may face a situation in the future when any other OS will require its own fix. Then you will have to make another duplicate function. Do you really want to do this for every next fix? Imagine what the user code will look like: a series of switch-cases. Now you have a great short API. It is transparent, and I, as a user, like it. There is no need to complicate it by duplicating the code.

NicoHood:
My whole point is, that this super duper rare case is not even worth adding. I assume not even 3 people will use it, so I'd rather make an example sketch with the mentioned fix (calling `x*2 + 32768`) instead of adding it to the library. I really struggle with myself to add this patch.

mdevaev:
Then let's just throw out this PR. I don't like the idea of duplicating the API, and I don't want to add code to your library that I won't use myself. My patch will remain without merge, and you can just refer to it in your documentation.

PR #315, opened 2021-08-29
Allow reducing gamepad buttons to 16 and removing Z-Axis easily

SukkoPera:
This patch does nothing by default, but adds a few configuration options for the above into HID-Settings.h.

### DIFF
--- a/src/HID-APIs/GamepadAPI.h
+++ b/src/HID-APIs/GamepadAPI.h
@@ -38,13 +38,18 @@ THE SOFTWARE.
 #define GAMEPAD_DPAD_LEFT 7
 #define GAMEPAD_DPAD_UP_LEFT 8
 
+#ifdef HID_ENABLE_32BUTTONS
+typedef uint32_t HID_Buttons_t;
+#else
+typedef uint16_t HID_Buttons_t;
+#endif
 
 typedef union ATTRIBUTE_PACKED {
 	// 32 Buttons, 6 Axis, 2 D-Pads
 	uint8_t whole8[0];
 	uint16_t whole16[0];
 	uint32_t whole32[0];
-	uint32_t buttons;
+	HID_Buttons_t buttons;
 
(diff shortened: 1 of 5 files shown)


NicoHood:
Why would you need this? Isn't more always better than less?

SukkoPera:
Not really, IMHO. Who needs 32 buttons? Which commercial controller has that many buttons? Implementing a 4-button controller and leaving 28 unused looks poor.

Besides that, a shorter USB report is probably ((very) slightly) quicker to transmit, which yields more CPU time.

NicoHood:
Flight simulator developers are more than happy for every button or axis they have.

The USB time is not really an argument for me.

SukkoPera:
Right, still I don't think that 99% of who is using this library has more than a few buttons. Anyway this PR has no impact on the default behaviour, it just adds a few settings that can be tweaked by advanced users, so flight simulator users will still be happy.

SukkoPera:
Upon testing on Windows 10, let me report that disabling the Z axis (i.e.: commenting out `HID_ENABLE_ZAXIS`) will result in the controller not being recognized.

On the other hand, reducing the buttons to 16 (`HID_ENABLE_32BUTTONS`) has no such impact and the controller still works fine.

If you want, I can remove the former from the PR.

NicoHood:
Honestly I do not think that this change is helpful for at least 1% of the library users. So I am not really into bloating the library with reducing its functionality for no good reason. I'd rather create a whole separate device called 'joystick' for example with even less buttons. I think I even had that before, but in general it made no sense, as you can just use the "better" device.

SukkoPera:
I don't think this would bloat the library, it would rather be unnoticed by most users IMHO, but that's ok, you're the boss :).

DRSDavidSoft:
@NicoHood I prefer to report the buttons that the project _actually_ have, so I'll be really pleased to be able to somehow define how many buttons are connected to the device :)

PR #316, opened 2021-08-29
Avoid unreferenced Gamepad instances to be created by the compiler

### LINKED ISSUE #266: Reduce number of pre-defined Gamepad instances
https://github.com/NicoHood/HID/blob/7e3741426137a099f72ed692dec38c44e1bf0ec3/src/SingleReport/SingleGamepad.cpp#L149-L152

I think it would be better to reduce the number of pre-defined single-report Gamepad instances to 2.

Having 4 does not make much sense IMHO as even on the most capable 32U4 there are only 3 usable endpoints, so besides barring the possibility to use the Gamepad with one or two different devices, the 4 instances cannot even show up all together.

Having two - or even a single one - is not a limitation, since additional devices can be declared in user sketches, if more are needed.

SukkoPera:
This seems to fix #266 upstream.

Since now only referenced Gamepad instances are actually created by the compiler, I took the liberty of adding Gamepad5 and 6.

### DIFF
--- a/src/HID-APIs/GamepadAPI.h
+++ b/src/HID-APIs/GamepadAPI.h
@@ -38,13 +38,18 @@ THE SOFTWARE.
 #define GAMEPAD_DPAD_LEFT 7
 #define GAMEPAD_DPAD_UP_LEFT 8
 
+#ifdef HID_ENABLE_32BUTTONS
+typedef uint32_t HID_Buttons_t;
+#else
+typedef uint16_t HID_Buttons_t;
+#endif
 
 typedef union ATTRIBUTE_PACKED {
 	// 32 Buttons, 6 Axis, 2 D-Pads
 	uint8_t whole8[0];
 	uint16_t whole16[0];
 	uint32_t whole32[0];
-	uint32_t buttons;
+	HID_Buttons_t buttons;
 
 	struct ATTRIBUTE_PACKED {
 		uint8_t button1 : 1;
@@ -65,6 +70,7 @@ typedef union ATTRIBUTE_PACKED {
 		uint8_t button15 : 1;
 		uint8_t button16 : 1;
 
+#ifdef HID_ENABLE_32BUTTONS
 		uint8_t button17 : 1;
 		uint8_t button18 : 1;
 		uint8_t button19 : 1;
@@ -82,6 +88,7 @@ typedef union ATTRIBUTE_PACKED {
 		uint8_t button30 : 1;
 		uint8_t button31 : 1;
 		uint8_t button32 : 1;
+#endif
 
 		int16_t	xAxis;
 		int16_t	yAxis;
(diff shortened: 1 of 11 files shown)


src/SingleReport/SingleGamepad.h
```diff
@@ -47,9 +47,15 @@ class SingleGamepad_ : public PluggableUSBModule, public GamepadAPI
...
+ * See #266 for details.
+ */
 extern SingleGamepad_ Gamepad1;
 extern SingleGamepad_ Gamepad2;
 extern SingleGamepad_ Gamepad3;
 extern SingleGamepad_ Gamepad4;
-
-
+extern SingleGamepad_ Gamepad5;
+extern SingleGamepad_ Gamepad6;
```
NicoHood, line 61: If I remember correct, only 3 will ever work, as the 32u4 only has 6 usb endpoints, of which 3 are used by the serial?

NicoHood:
Does it **seem** to fix, or does it **really** fix the issue?

src/SingleReport/SingleGamepad.h
SukkoPera: Correct, but they turn up handy if the CDC serial gets disabled. Since with this modification their presence won't cause any overhead, I thought I'd throw them in.

SukkoPera:
It fixes the issue on a project of mine. I need to test more to say it definitely does with confidence. Just thought I'd share the patch in the meantime, so that maybe other people could test on their own.

SukkoPera:
Just noticed an odd side-effect: Gamepad1 no longer corresponds to the first gamepad detected by the system, sigh.

NicoHood:
I think the order of Gamepad creation could be related to: 

* First use in the code
* Compiler sorting, possibly by name

So you could try to change the name of the cpp files maybe. Another idea would be to add special attributes to the class to change its construction order. I am not sure if that is possible, but I remember that there could be options like that.

Is it even important to have them in the correct order?

SukkoPera:
I thought the same, but the first use in my code is in an array which is populated in order.

Files and instances are named in ascending order so I see no way that could lead to the revers order.

I think users would expect Gamepad1 to be the first controller, I think we should preserve this behavior.

NicoHood:
I dont know, but I bet you will find it out.

SukkoPera:
More testing confirms that this solves the initial issue, but still I can't understand why the order of controllers gets reversed.

RevHarryPowell:
Sorry to revive this old conversation, but I’m having trouble implementing SukkoPera’s fix in my code, which currently has only 2 HID gamepads. The only way I can get rid of the third is by deleting the extra SingleGamepadX.cpp files and removing the lines in SingleGamepad.h. I’d like to be able to dynamically change the number of HID gamepads generated. Is there a way to accomplish this?

fermino:
Hi! First of all, @NicoHood I know that maintaining a project isn't easy, so thanks a lot for this library!

Is there any chance to have this PR merged? I am experiencing the same issues as in #266, and although creating instances in the lib files is common practice in the arduino environment, it get's kind of annoying when you try to dive a bit deeper. 

For what I've been reading the order of compilation is not defined by the C standard, but if there's a GCC behavior we could try to hack around that to make it work for most of the people.

PR #329, opened 2021-11-15
Add #ifndef HID_CUSTOM_LAYOUT_HIDE_PRAGMA_MESSAGE

fxkr:
Currently, on every build using this library, this message is printed:

```
In file included from .pio/libdeps/promicro/HID-Project/src/SingleReport/../HID-APIs/KeyboardAPI.h:29:0,
                 from .pio/libdeps/promicro/HID-Project/src/SingleReport/../HID-APIs/DefaultKeyboardAPI.h:27,
                 from .pio/libdeps/promicro/HID-Project/src/SingleReport/BootKeyboard.h:30,
                 from .pio/libdeps/promicro/HID-Project/src/HID-Project.h:50,
                 from src/input.h:9,
                 from src/main.cpp:4:
.pio/libdeps/promicro/HID-Project/src/SingleReport/../HID-APIs/../KeyboardLayouts/ImprovedKeylayouts.h:57:21: note: #pragma message: Using custom layout for keyboard modules
     #pragma message "Using custom layout for keyboard modules"
```

(Or similar, depending on #define's and whether Arduino IDE or PlatformIO is used to build.)

I like to try and keep my errors/warnings/messages count at 0. This makes it easier to see if any in development changes have introduced new warnings (visually by length + color of build output), making for a more pleasant development experience. These messages are printed in scary yellow (PlatformIO) or red (Arduino IDE), and in Arduino IDE they're especially annoying since the build log panel is usually a very small portion of the screen.

Of course personally I would prefer removing the message entirely, but I assume you would not be ok with that (someone must've added it for some reason. maybe it reduces support load or something), so in this PR I'm proposing to at least add a way to disable the message via another #define. Since this is meant for people who're pretty sure they know what they want, I only added the #ifndef to the branch where a layout is already specified. And the default behavior is unchanged.

Does this seem like an acceptable compromise?


src/KeyboardLayouts/ImprovedKeylayouts.h
```diff
@@ -53,7 +53,9 @@ enum KeyboardLeds : uint8_t {
     #define LAYOUT_US_ENGLISH
     #pragma message "Using default ASCII layout for keyboard modules"
```
NicoHood, line 54: If we want to go this way, this pragma also needs a wrapper.

NicoHood:
I am not sure if that is a useful idea. I'd expect that the compiler has a flag to turn off pragma messages.

fxkr:
I don't think there is a way to disable `#pragma message`. I searched; the only thing that came up is this StackOverflow post https://stackoverflow.com/questions/31669021/how-to-silence-pragma-messages-from-gcc that also proposes #ifdef.

Even if there was a compiler option to disable pragma message:

1. That's difficult/impossible to integrate into the build system (especially in Arduino IDE, where it is impossible to override build options in project specific settings. boards.txt is system wide. Then I could as well patch it out of HID-Project directly.)
2. If it was done globally (which, I believe, would be the only technically possible in Arduino IDE), then one wouldn't know if anything else is using them to raise warnings if something is wrong. The goal here is _not_ to "just make it shut up at all cost" (`&>/dev/null` style), but to reduce noise when all is right.

I believe HID-Project is the only library that I am aware of (at home or at work) that prints something as part of its regular (successful, no warnings) build.

src/KeyboardLayouts/ImprovedKeylayouts.h
fxkr: I explained why I did it like this above, but I do not actually care about whether one or both branches get the ifndef, so I have updated the PR to add the 2nd wrapper as you said.

fxkr: (I have also changed it from HID_CUSTOM_LAYOUT_HIDE_PRAGMA_MESSAGE to HID_HIDE_LAYOUT_PRAGMA_MESSAGE so that the same define properly describes both #pragma messages - since the 1st pragma message is regarding the default layout, not a custom layout.)

NicoHood:
FastLED also uses the pragma to print the version.

Any thoughts from other community members?

PR #332, opened 2021-11-27
Add Japanese keylayout

HisashiKato:
Hello.
Thanks for the great library.

I have created a "Japanese keylayout".
I'm the only one who checked it, so it may be wrong.

I'm not very good at English.
I am a beginner in programming and GitHub.
If I am rude, please forgive me.

(This was translated from Japanese to English using a translation website.)


NicoHood:
Thanks! I will just accept it, if somebody reports an error, we can fix it later!

PR #342, opened 2022-01-10
fix missing return in function requiring return value!

ladyada:
this was throwing a compilation error on our SAMD core (we require functions to return values if they say they will :) functionality should now be the same on any/all compilers


NicoHood:
Wow, I am wondering why the initial patch did work. Thanks!

ladyada:
*some* compilers will automatically add a 'return' for the last statement in a function...but not all :)

could you do a version bump and release? that will make it easier for folks to get the latest updates from the library manager!

PR #393, opened 2022-12-26
Adding Horizontal Scroll Mouse Wheel Support

WaGi-Coding:
Might need to get checked more, but seems everything still works fine with my tests.

With this, you cann pass a 4th parameter for horizontal scroll value in Mouse and AbsoluteMouse move/moveTo

Please review

### DIFF
--- a/src/HID-APIs/AbsoluteMouseAPI.hpp
+++ b/src/HID-APIs/AbsoluteMouseAPI.hpp
@@ -73,7 +73,7 @@ void AbsoluteMouseAPI::click(uint8_t b){
 	moveTo(xAxis, yAxis, 0);
 }
 
-void AbsoluteMouseAPI::moveTo(int x, int y, signed char wheel, signed char hWheel){
+void AbsoluteMouseAPI::moveTo(int x, int y, signed char wheel){
 	xAxis = x;
 	yAxis = y;
 	HID_MouseAbsoluteReport_Data_t report;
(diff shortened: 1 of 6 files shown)


NicoHood:
I did not test this, but just looking at it, it seems fine to me. However I am not sure if this will break some other devices, as the descriptor is now more complex, not sure.

WaGi-Coding:
So far it all works fine for me. Playing around with 2 Rotary Encoders simulating a mouse. With held down buttons, while still being able to move, so you can drag stuff around or draw & an optional setting to simulate a held press by holding the button for half a second, as it could be hard to hold down and move with that setup ^^ Also with option to switch between modes (mouse, scroll, volume/seekbar) & a TFT for configuring that to ones likings.

Maybe implementing a whole keyboard you can use with the encoders and the TFT, to test it more in combination with other devices.


Would be great if others could test their existing, more complex projects with that pull request fork to make sure.

Greetings and happy new year

darylwright:
Hi @WaGi-Coding, I sent you a pull request on your fork. I suggest the horizontal scroll functionality be put in a separate method so as not to break existing projects by changing the API. This should address @NicoHood's concerns. Please let me know what you think of these changes.

WaGi-Coding:
Thanks @darylwright, I merged it into my forks master/base branch and left the old code in another branch, because i refer to it in a tutorial. Many thanks!

NicoHood:
If the function has a default value, why would it break?

src/HID-APIs/AbsoluteMouseAPI.hpp
```diff
@@ -85,12 +85,18 @@ void AbsoluteMouseAPI::moveTo(int x, int y, signed char wheel, signed char hWhee
...
 }
 
-void AbsoluteMouseAPI::move(int x, int y, signed char wheel, signed char hWheel){
-	moveTo(qadd16(xAxis, x), qadd16(yAxis, y), wheel, hWheel);
+void AbsoluteMouseAPI::move(int x, int y, signed char wheel){
+	moveTo(qadd16(xAxis, x), qadd16(yAxis, y), wheel);
+}
+
+void AbsoluteMouseAPI::scroll(signed char wheel, signed char hWheel){
+	HID_MouseAbsoluteReport_Data_t report;
```
NicoHood, line 96: Did you test this? Doesnt this set the cursor to a fixed position, maybe 0,0?

darylwright:
As I mentioned in @WaGi-Coding's fork, my changes were not necessary and those prior to my involvement were satisfactory. Apologies for the inconvenience.

